PR #4, opened 2015-11-24
Feature dashboard

roobottom:
Patterns for the dashboard


blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.less
```diff
@@ -0,0 +1,32 @@
+//header
+.ff_container-dashboard-section__header {
+    padding: @ff_size_control_vertical;
```
dkhuntrods, line 3: I think we need a different var setup. In the new style guide, there's three situations - the 'Box' has 10px 20px, the tab box has 15px 20px 20px 20px, the filter box has all 15px. So probably a var for each value, and three shortcut vars that combine them in each situation?

blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.less
```diff
@@ -0,0 +1,32 @@
...
+    background-color: @ff_colour_control_bg;
+    border: 1px solid @ff_colour_control_border;
+    border-top-left-radius: @ff_size_control_rounding;
+    border-top-right-radius: @ff_size_control_rounding;
+}
+
+.ff_container-dashboard-section__title {
+    font-size: 1rem;
```
dkhuntrods, line 11: Do we need rem-fallbacks, or were we going to try something else for font sizes?

roobottom, line 11: Yes, we do need rem fall-backs. Can we write a common mixin for this task do you think?

PR #5, opened 2015-11-25
Feature task listing

sd-92:


simonedeluca-firefly:
This was a mistake!

blocks/core/_shared/settings.less
```diff
@@ -16,11 +16,19 @@ and shared between all themes */
 @ff_colour_label_pink:								#FF97CC;
 @ff_colour_label_purple:							#9A63CE;
 
+<<<<<<< HEAD
```
dkhuntrods, line 19: Conflict markers are still in here, can you take a look?

blocks/core/ff_module/ff_module-filter/ff_module-filter.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+$(function(){ 
+	$("[data-control='filter-dropdown']").click(function(){
+		var filterDropdown = $(this),
+		      filterIcon = filterDropdown.find("[data-icon='filter-dropdown-icon']"),
+		      filterContent = $("[data-content='filter-content']");
```
dkhuntrods, line 7: Hi Simone, have you tested this with multiple filters on a page? It looks to me like this would select all other `data-content='filter-content'` elements on the page.

PR #8, opened 2015-11-26
Feature task listing

sd-92:


blocks/core/_shared/settings.less
```diff
@@ -19,6 +19,7 @@ and shared between all themes */
 
 @ff_colour_control_bg: @ff_colour_quiet;
 @ff_colour_control_bg_selected: darken(@ff_colour_quiet, 10%);
+@ff_colour_control_bg_lighter: lighten(@ff_colour_quiet,5%);
```
roobottom, line 22: hmm. Not sure about this. @dkhuntrods what do you think? Make a new variable here, or in the module?

dkhuntrods, line 22: @roobottom I think small variations of the base settings like this should belong to the module, yeah.

blocks/core/ff_module/ff_module-filter/ff_module-filter.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+$(function(){ 
+	$("[data-control='filter-dropdown']").click(function(){
+		var filterDropdown = $(this),
+		      filterIcon = filterDropdown.find("[data-icon='filter-dropdown-icon']"),
+		      filterContent = filterDropdown.next();
```
dkhuntrods, line 7: I don't think it's a good idea to use .next() in the module js, as it ties the code to the structure of the html, which may change. Instead, the initiating data-\* declaration should be at the parent element of all effected elements, and each essential element should have the data-\* attribute that indicates its function and also its relation to the overall module. @roobottom perhaps we should come up with a js naming convention too? Or invent jsBEM? :)

roobottom, line 7: @dkhuntrods agree. It's worth having a chat with Michael as he and I had a chat around naming of `data` elements on Friday. He's got some good ideas on this.

PR #12, opened 2015-11-30
Feature task listing

roobottom:
Changed:
- The way `js` attaches itself to DOM elements
- Removed an extra variable from the core that didn't need to be there

Added
- Styles for Melody


blocks/core/ff_module/ff_module-filter/ff_module-filter.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+$(function(){ 
+	$("[data-ff-action='filter-control']").click(function(){
+		var filterDropdown = $(this),
+		      filterIcon = filterDropdown.find("[data-ff-icon='filter-dropdown-icon']"),
+		      filterContent = filterDropdown.closest("[data-ff='filter']").find("[data-ff='filter-content']");
```
dkhuntrods, line 7: @roobottom just curious, does it need the `closest("[data-ff='filter']")` bit, is that doing something in a case I can't see just by looking at the code?

roobottom, line 7: Probably not, it's just my lack of `js` skills. I want to find the `content` in a reliable way, so thought it best to transvers up the tree to a known node, then `find()` within that.

Is there a better way of doing this?

dkhuntrods, line 7: That's probably fine, usually I'd try to set the handler on the module itself and use delegation (have fewer listeners as things get more complex, which is always good) but in this case this is probably the simplest way.

PR #19, opened 2015-12-16
Common feature icons

sd-92:


dkhuntrods:
Please add icons/optimized_svgs into .gitignore :)

PR #25, opened 2015-12-18
Make first form step active on init

mwilliamson-firefly:
As the comments says: this is an ugly hack to get the first step showing by default. Either the modified method should be renamed or some other change is required, but I'm not sure what structure you'd prefer.

### DIFF
--- a/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
+++ b/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
@@ -120,6 +120,9 @@ function getSwapStatesMethod($root, options) {
 function setClickHandler($root, options) {
     var linkSel = '[' + options.linkSelBase + ']';
     $root.on('click', linkSel, getSwapStatesMethod($root, options));
+    // FIXME: this is an ugly hack to get the first step showing by default
+    // Either this method should be renamed or some other change is required
+    $root.find(linkSel).get(0).click();
 }
 
 module.exports = {


dkhuntrods:
@mwilliamson-firefly Is there a reason why the class can't initially be set on page render?

mwilliamson-firefly:
@dkhuntrods I did the easiest thing that seemed to ensure that the first step was activated in the same way as if were clicked. I'd suggest that, for consistency, having a function that can activate steps to ensure that activating the first step on load is the same as clicking the first step. That's effectively what this horrible hack does, except it doesn't directly expose said function (because I didn't want to impose a structure on code I don't own)

dkhuntrods:
@mwilliamson-firefly Oh yes, of course it should have a method for that (though in this case, very intentionally, all the code does is add and remove classes). I'll add one in; I do think it should be activated separately and intentionally (in case we want to load the page to a specific step at some point) rather than activating the first one by default - I'm guessing that's how you're thinking too?

mwilliamson-firefly:
@dkhuntrods It feels to me like you want to select a step when the steps are initialised, and that a default of the first step is correct most of the time. If the caller needs to ever start on a different step, then I'd say let them specify that on initialisation, but no need to implement it until we actually need it. In other words, something like:

```
initialiseFormSteps({container: formStepsElement}); // defaults to first step
initialiseFormSteps({container: formStepsElement, activeStep: "attachments"}) // if we ever need, but I'd say leave it out for now (names are arguably more resilient that indices, but not implementing this avoids having to make the decision anyhow)
```

PR #27, opened 2015-12-22
Common feature tabs update

dkhuntrods:
Updating tabs js to add next(), last(), and immediate activation of first tab.
@mwilliamson-firefly @roobottom

### DIFF
--- a/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
+++ b/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
@@ -144,18 +144,19 @@ function getTabHandler($root, options) {
         selectedIndex = testBounds(index, $triggers.length);
         var trigger = $triggers.get(selectedIndex);
         if (trigger) setState(trigger);
+        return main;
     }
 
     function setInitialTab() {
-        setActiveTab(selectedIndex);
+        return setActiveTab(selectedIndex);
     }
 
     function next() {
-        setActiveTab(selectedIndex + 1);
+        return setActiveTab(selectedIndex + 1);
     }
 
     function last() {
-        setActiveTab(selectedIndex - 1);
+        return setActiveTab(selectedIndex - 1);
     }
 
     main.handleClick = handleClick;
--- a/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
+++ b/blocks/core/_lib/ff-tabs/_ff-tabs-core.js
@@ -14,7 +14,8 @@ var _options = {
     completedClassSuffix: '--is-complete',
     visitedClassSuffix: '--is-visited',
     isComplete: isComplete,
-    canAdvance: canAdvance
+    canAdvance: canAdvance,
+    selectedIndex: 0
 };
 
 function isComplete($nextLink, $nextContent, $currentLink, $selectedContent) {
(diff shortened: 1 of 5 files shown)


blocks/core/_lib/ff-tabs/_ff-tabs-core.js
```diff
@@ -113,16 +122,61 @@ function getSwapStatesMethod($root, options) {
...
+    }
+
+    function last() {
+        return setActiveTab(selectedIndex - 1);
+    }
+
+    main.handleClick = handleClick;
+    main.setActiveTab = setActiveTab;
+    main.next = next;
+    main.last = last;
```
mwilliamson-firefly, line 165: Perhaps `previous` instead of `last`? Last is normally the final element.

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
```diff
@@ -113,16 +122,61 @@ function getSwapStatesMethod($root, options) {
...
-    $root.on('click', linkSel, getSwapStatesMethod($root, options));
+    function testBounds(value, length) {
+        if (value < 0) {
+            return length - 1;
+        } else if (value >= length) {
+            return 0;
+        } else return value;
+    }
+
+    function setActiveTab(index) {
```
mwilliamson-firefly, line 142: I think this is fine for internal use (with last/next), but is it necessary to expose it as part of the interface? If so, I'd suggest that a function that identifies tab by some sort of name would be less brittle than by index.

roobottom:
@mwilliamson-firefly This is above my pay grade. Would you kindly review this?

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
dkhuntrods, line 142: Sorry, which part do you mean?
We do need methods that simply advance the tabs, for the set-a-task form step flow, so I do think we need next and (last/previous)

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
dkhuntrods, line 165: Sure

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
mwilliamson-firefly, line 142: I agree we need previous/next, I'm not sure that we need to expose `setActiveTab`.

mwilliamson-firefly, line 142: Or `handleClick`, come to think of it.

dkhuntrods, line 142: Ah, right. Not strictly no, but had gotten into the habit of exposing a fairly full API for testing (which haven't had a chance to do this time around, but would like to add them in at a later point, and since they're already there..?)

mwilliamson-firefly, line 142: I tend to only add (and test) things that are required by production code, but I don't think I'm really in a position to comment on how you're developing your code! I was just thinking more about what gets exposed to the user -- it's easier to expose more things, but harder to take them away/change them.

PR #29, opened 2016-01-05
Form preview dd attributes

bwfox:
Allow pattern library consumers to specify data attributes to assign to form preview list items.

### DIFF
--- a/blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.xsl
+++ b/blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.xsl
@@ -21,9 +21,9 @@
             </xsl:call-template>
             </dt>
             <dd class="ff_module-form-preview__list__data">
-                <xsl:if test="@data-ff-preview-for">
+                <xsl:if test="@preview-for">
                     <xsl:attribute name="data-ff-preview-for">
-                        <xsl:value-of select="@data-ff-preview-for"/>
+                        <xsl:value-of select="@preview-for"/>
                     </xsl:attribute>
                 </xsl:if>
             <xsl:choose>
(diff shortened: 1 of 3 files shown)


blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.xsl
```diff
@@ -20,7 +20,13 @@
...
             </xsl:call-template>
             </dt>
-            <dd class="ff_module-form-preview__list__data">
+            <dd>
+                <xsl:attribute name="class">
+                    <xsl:text>ff_module-form-preview__list__data</xsl:text>
+                    <xsl:if test="@data-ff-attr">
+                        <xsl:text> data-ff-</xsl:text><xsl:value-of select="@data-ff-attr"/>
+                    </xsl:if>
+                </xsl:attribute>
```
mwilliamson-firefly: I might be misunderstanding what's going on here, but wouldn't we want to have those as separate attributes rather than as classes? In other words, isn't the HTML we want to end up with something like:

``` html
<dd class="ff_module-form-preview__list__data" data-ff-preview-for="some-locator">
```

(Assuming the purpose of the change is to point the summary field to wherever it should grab its value from)

bwfox:
@mwilliamson-firefly So something like that?

mwilliamson-firefly:
Seems sensible. I might be tempted to rename the attribute being passed in to match the attribute being generated i.e. `preview-for` to make it clearer exactly what it's a locator for.

dkhuntrods:
@bwfox would you mind updating the .md and .xml files to reflect this addition?

blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.xsl
```diff
@@ -21,6 +21,11 @@
             </xsl:call-template>
             </dt>
             <dd class="ff_module-form-preview__list__data">
+                <xsl:if test="@data-ff-preview-for">
```
mwilliamson-firefly: Ah, sorry, I wasn't very clear: I meant just `@preview-for` since we're already in the context of Firefly. Not too fussed about it though.

PR #33, opened 2016-01-06
Fix editor styles in planner lesson plans

mwilliamson-firefly:
The `ff_util-prose` class was interfering with the editor styles, so I've removed it. This is probably a bit drastic, but I wasn't sure how we want to handle the interaction between these styles. Feel free to accomplish the same thing with something less hacktastic!


roobottom:
@mwilliamson-firefly I agree that this causes problems, we're essentially styling this twice at the moment. We should readdress how we're doing this after Bett

PR #34, opened 2016-01-06
Allow some containers to be used as templates

mwilliamson-firefly:
This tweaks some of the containers so that they can be used as containers from Firefly.

@roobottom I think you mentioned that containers were meant to be copy and pasted instead of used as templates. Is there a reason why tweaking them as in this pull request is a bad idea?

### DIFF
--- a/blocks/core/ff_container/ff_container-page-header/ff_container-page-header.xml
+++ b/blocks/core/ff_container/ff_container-page-header/ff_container-page-header.xml
@@ -1,3 +1,3 @@
-<page-header title="Whatever">
+<page-header title="Page title">
     <controls title="Whut"/>
 </page-header>
--- a/blocks/core/ff_container/ff_container-page-header/ff_container-page-header.xsl
+++ b/blocks/core/ff_container/ff_container-page-header/ff_container-page-header.xsl
@@ -2,10 +2,12 @@
 	<xsl:param name="data" />
     
     <div class="ff_container-page-header">
-        <h1 class="ff_container-page-header__title">Page Title</h1>
-        <div class="ff_container-page-header__controls">
-            <div class="crate_util-block">Steps, or whatever</div>
-        </div>
+        <h1 class="ff_container-page-header__title"><xsl:value-of select="$data/page-header/@title"/></h1>
+        <xsl:if test="$data/page-header/controls">
+            <div class="ff_container-page-header__controls">
+                <div class="crate_util-block">Steps, or whatever</div>
+            </div>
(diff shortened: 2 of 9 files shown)


dkhuntrods:
@mwilliamson-firefly Did you test these within the pattern library? They're used with a variety of data for testing, and the util blocks mark out the section areas (the pink parts in the patternlib) 
If you can replicate the existing content with all it's test pages using these changes, great. I suppose the  `crate_util-block` spans or divs could be added within the `.md` directly?

mwilliamson-firefly:
@dkhuntrods Not yet, I wanted to check whether there was some reason this approach didn't work first. (Also, I seem to get errors when trying to serve the pattern library, but I can look into those)

roobottom:
What's the intention behind this? Sorry, struggling to understand what this does.

mwilliamson-firefly:
@roobottom Based on the meeting we had about the pattern library, I think you said that we had to copy and paste the HTML for containers. This let's you use the containers as templates (the same as the modules), so there's no more copy and paste.

If you look at the [corresponding pull request in Firefly](https://github.com/fireflylearning/firefly-webapp/pull/1336), you can see how they're used e.g.

``` xsl
    <xsl:variable name="page-content">
      <content>
        <div data-ff-tabs="">
          <xsl:call-template name="ff_module-tabs-navigation" >
            <xsl:with-param name="data" select="msxsl:node-set($tabs)" />
          </xsl:call-template>
          <xsl:call-template name="ff_container-tabs-content">
            <xsl:with-param name="data" select="msxsl:node-set($tabs)" />
          </xsl:call-template>
        </div>
      </content>
    </xsl:variable>

    <xsl:call-template name="ff_container-page">
      <xsl:with-param name="data" select="msxsl:node-set($page-content)"/>
    </xsl:call-template>
```

mwilliamson-firefly:
@dkhuntrods I've amended the commit to include appropriate changes to the XML in the pattern library.

dkhuntrods:
@mwilliamson-firefly I've just pulled and run your branch --  that's fantastic! I suspected it should be possible to do something like this but my xsl-fu isn't 100% as yet.

Re. the errors - your branch built without errors when I ran it just now. I've got pretty aggressive cacheing going on, so if you get a persistent error it's worth running a `clean` and rebuild.

@roobottom If you're happy with this I'll merge?

roobottom:
@mwilliamson-firefly @dkhuntrods Yes, If you're happy, I'm happy!

PR #41, opened 2016-01-08
Use title provided by planner rather than generating one

bwfox:


blocks/core/ff_module/ff_module-date-stepper/ff_module-date-stepper.xsl
```diff
@@ -3,15 +3,7 @@
...
-            <xsl:choose>
-                <xsl:when test="$data//planner/@type = 'day'">
-                    <span class="ff_module-date-stepper__title--type">Today</span>
-                </xsl:when>
-                <xsl:when test="$data//planner/@type = 'week'">
-                    <span class="ff_module-date-stepper__title--type">Week</span>
-                </xsl:when>
-            </xsl:choose>
-            <span class="ff_module-date-stepper__title--date"><xsl:value-of select="$data//planner/@date"/></span>
+            <span class="ff_module-date-stepper__title"><xsl:value-of select="$data//planner/@title"/></span>
```
bwfox: @roobottom Do I still want this span or should I move the class into the `h3`?

roobottom: @bwfox This still needs to be two separate items as the design calls for the day-name to be bold. Se we need to retain the `--type` and `--date` `<span>`'s

dkhuntrods:
@bwfox @roobottom  What's the status of this? Should this PR be closed?

bwfox:
I was going to tweak it, since we don't want type to be either `Day` or `Week`, but for day view be a day of the week or a relative day name, e.g., "Tomorrow".

PR #43, opened 2016-01-08
Extracting date-picker from planner picker behaviour

AdamLynamFirefly:
Added support for date picker to set value on provided input id


roobottom:
@AdamLynamFirefly LGTM. If you don't need this branch, please delete it.

PR #45, opened 2016-01-11
Fixing premature update of results and query state

dkhuntrods:
Updating recipient picker to prevent premature setting of state to prevent 'No results' message firing before results returned @mwilliamson-firefly


roobottom:
@dkhuntrods merged. Please delete branch if not needed.

PR #46, opened 2016-01-11
Feature set a task

dkhuntrods:
Refactored formsteps and tabs patterns to be more congruent 

**NOTE**: some minor but potentially breaking changes here
- Tab and formsteps now use same convention for state:
  - ~~`<(tab|step) active="true | false">`~~
  - `<(tab|step)  state="is-active | is-current | is-etc">`
- Some elements renamed for consistency

Also:
- Built two new test patterns, one for the formsteps and one for the set-a-task page with picker components and overlay.
- Converted many container patterns to reusable xsl templates.


roobottom:
@dkhuntrods any reason this is a pull request, and not just a merge-in?

dkhuntrods:
@roobottom mainly since there were breaking changes, to document for all to see.

PR #51, opened 2016-01-11
Adding bordered box container

AdamLynamFirefly:


dkhuntrods:
@AdamLynamFirefly would you mind having the content coming from the markdown file rather than hard coded in the xml? That way we can use it as a real container elsewhere in the pattern lib for testing and such.

roobottom:
@AdamLynamFirefly didn't we say we were going to use `item-repeater` for this? I'm not convinced that `borderd-container` is semantic, as it describes the style rather than functionality.

AdamLynamFirefly:
I asked about that on the task details PR. item-repeater has rounded borders doesn't it?

If we are going to talk semantic's it also doesn't sound like a suitable component either. :P

roobottom:
@AdamLynamFirefly You're right, but I'm not suggesting that the repeater component is the way to go long term, I'm just saying lets use this for now, rather than introducing a new pattern that we've not really fully thought about. I'm sure the rounded corners will be ok for bett, too.

PR #56, opened 2016-01-15
Display none on validation message

AdamLynamFirefly:
Stops the page from adjusting when the message is immediately removed from the dom after loading the page


mwilliamson-firefly:
This seems fine to me, not sure whether a designer is supposed to look at this though?

AdamLynamFirefly:
Good point @mwilliamson-firefly, @roobottom any chance you could take a look / get somebody to take a look at this before it is merged?

roobottom:
LGTM, although it will now _not_ show in the pattern lib by default. Do you think this is OK @dkhuntrods ?

PR #66, opened 2016-02-01
Allow export path to be set as a command line argument

mwilliamson-firefly:


dkhuntrods:
@mwilliamson-firefly  I'll have to rewrite this for the new setup, but will merge here anyways.

PR #69, opened 2016-02-03
Handle pattern-library.json not already existing

mwilliamson-firefly:


gulpfile.js
```diff
@@ -266,7 +266,13 @@ gulp.task('build', ['xslt', 'css', 'assets', 'js', 'icons:copy']);
  *********************************************/
 gulp.task('export', ['export:blocks', 'export:less', 'export:js', 'export:icons', 'export:assets'], function(cb) {
     var infoPath = path.join(config.exportPath, "pattern-library.json");
-    unlink(infoPath).then(function() {
+    unlink(infoPath).catch(function(error) {
+        if (error.code === "ENOENT") {
+            return null;
+        } else {
+            return bluebird.reject(error);
```
karim-firefly, line 273: What happens when this returns `null` versus rejecting the promise with the specific error as reason?

mwilliamson-firefly, line 273: Returning `null` means resolve the promise with `null` (where resolve means successful fulfill the promise), whereas rejecting the promise means fulfilling the promise with a failure rather than a success.

PR #73, opened 2016-02-04
ff_module-task - href fix

sd-92:
@href doesn’t exists it’s defined in the xml as link_href

### DIFF
--- a/blocks/core/ff_module/ff_module-task/ff_module-task.xsl
+++ b/blocks/core/ff_module/ff_module-task/ff_module-task.xsl
@@ -14,7 +14,7 @@
             </xsl:if>
         </xsl:attribute>
         <dt class="ff_module-task__item ff_module-task__item--title">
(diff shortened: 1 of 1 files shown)


blocks/core/ff_module/ff_module-task/ff_module-task.xsl
```diff
@@ -14,7 +14,7 @@
             </xsl:if>
         </xsl:attribute>
         <dt class="ff_module-task__item ff_module-task__item--title">
-            <a href="{$data/notice/@href}" class="ff_module-task__link"><xsl:value-of select="$data/notice/htmlMessage"/></a>
+            <a href="{$data/notice/@link_href}" class="ff_module-task__link"><xsl:value-of select="$data/notice/htmlMessage"/></a>
```
roobottom, line 17: @simonedeluca-firefly How does this get called — via a variable, or is it fed the XML from the Firefly page directly?

PR #74, opened 2016-02-05
Add profile response button

mwilliamson-firefly:
This adds in the profile response button that Danielle was working on, as well as allowing `ff_container-item-repeater` to be used as a React component. Further styling changes are probably needed, but this needs merging before the work done for the responses page can be merged in.


blocks/core/ff_module/ff_module-tabs/ff_module-tabs.md
```diff
@@ -0,0 +1 @@
+# TODO: Tabs API
```
bwfox, line 1: Are we okay with merging this in?

mwilliamson-firefly, line 1: Err, not sure what Danielle had in mind when she wrote it.

bwfox:
LGTM

PR #79, opened 2016-02-16
Dkh feature inline block js

dkhuntrods:
Adding automatic inclusion of block js files to the patternlib pages so that it is no longer necessary to manually update and maintain (or include) the shared/index.js entry point file.
See changelog for more details.
@roobottom


roobottom:
Tested, seems to work OK.

PR #81, opened 2016-02-17
Fix rounded corners on filter-dropdown

sd-92:
I decided to fix it in the container item repeater as it makes more sense.


dkhuntrods:
Are you sure overflow:hidden here won't have any unexpected repercussions?
Actually, I've fixed this on a branch I'm working on with overflow:auto which also works and might be a bit more robust, so if it's not needed urgently we can close this and merge in when I PR my branch if that's OK.

sd-92:
That sounds good. It is not urgent so we can close this and merge your branch later!! Thanks.

PR #82, opened 2016-02-17
ff_module-progress bar "sausage" fix

sd-92:
Fixed the logic that adds the --lonely modifier


sd-92:
I've tested this with all the possibilities:

screen shot 2016-02-17 at 15 58 53
screen shot 2016-02-17 at 16 00 37
screen shot 2016-02-17 at 16 01 11
screen shot 2016-02-17 at 15 59 15

PR #85, opened 2016-02-18
Container filter dropdown responsive updates

sd-92:
I am not sure whether or not this was the right way of doing it. Comments are welcome!


dkhuntrods:
Unfortunately I don't have a lot of time to review in-depth right now, but one quick question-- should the file `.icons/build/png/ff_icon-page-forward-open-lightblue.png` have changed? I'll try and have a proper look at the rest this evening!

sd-92:
That's strange, I didn't directly change that file. What I did is run the gulp icons command not sure why it removed it?

dkhuntrods:
If the colour definition (in the gulpicon config) for light-blue was changed or removed then the generated file would change. Can you check if that was modified, and revert if it wasn't intentional, and regenerate the icons? If it was intentional, and it's a new thing you're adding, can you create a new colour definition instead of changing the existing one? Thanks!

sd-92:
That is now updated with the fix to the light blue colour!

blocks/core/ff_container/ff_container-filter-dropdown/ff_container-filter-dropdown.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+$(function(){ 
+	$("[data-ff-action='filter-toggler']").click(function(){
+		var filterTogglerText = $("[data-ff='filter-content']").is(':visible') ? 'Filter this list' : 'Hide filters' ;
```
dkhuntrods: It might be a good idea to make the visibility state a bit more explicit, in order to prevent the toggles getting out of sync; something like (pseudo code)

```
var $content = $(data-filter-content),
$icon = $(data-toggler-icon),
$text = $(data-toggler-text);

var isVisible = $content.is(:visible); 
var newText = isVisible ? 'Filter this list' : 'Hide filters';
if (isVisible) {
  $content.hide();
  $icon.removeClass(visibleClass).addClass(hiddenClass);
else {
  $content.show();
  $icon.removeClass(hiddenClass).addClass(visibleClass);
}
$text.text(newText);
```

sd-92:
@dkhuntrods thanks for the comment. I did that at first but then I thought that writing less code was "better". I agree that it's more readable this way and might prevent the toggle from acting funny! (I've updated it)

PR #86, opened 2016-02-19
Add module link button

aegirfirefly:
Pretty much a direct copy of the module button but for 'a' tags. Should look the same (though will expand to fill available width)


dkhuntrods:
Thanks Aegir! Have the md docs been changed to reflect the slightly different usage? Ie. you'd pass in a <link-button> rather than a <button>, if I'm reading correctly?

aegirfirefly:
No, good point. I also have to fix a layout issue with this and with the module_button on narrow viewports, so might close this and do that first.

PR #88, opened 2016-02-22
Feature task details response

dkhuntrods:
@mwilliamson-firefly I know there's a lot here, but if you could cast an eye quickly and make sure it's looking in the right direction at least!
There are tests for all the new components. They're based on the state of the stubs from last week, so there's a bit of updating to do, but I'd like to get things merged if possible before doing much more.


blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
```diff
@@ -1,9 +1,32 @@
...
-# Here be a profile-response button
+# Profile response button
 
 ## Markup 
 ```
 <div data-ff_module-profile-response-button=""/>
 ```
+### Props
 
+- `onSelect`: Function
```
mwilliamson-firefly: Perhaps worth also stating the signature of the function i.e. `onSelect()` (or at least I'd been treating it as zero-arg callback -- perhaps worth actually implementing it as such to make sure callers don't accidentally depend on the event being passed through)

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
```diff
@@ -1,9 +1,32 @@
...
+- `uiState`: String; css defined for `is-selected`, `is-updated` (default), `is-reviewed` in Melody
```
mwilliamson-firefly: Are the states specific to Melody? It feels like the states shouldn't be specific to any particular theme.

Perhaps `status` is a better name than `uiState`? Perhaps worth asking Viv what collective noun we use for the different states that the button can be in and using that.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
```diff
@@ -1,9 +1,32 @@
...
+- `guid`: String; unique id
```
mwilliamson-firefly: Is the GUID only used for `data-guid`? What's `data-guid` used for?

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly: Oh, just saw the status field a bit later. Hmm, it feels like `uiState` is combining two separate notions that the component should be responsible for dealing with, specifically:
- whether or not it's selected (which feels like UI state)
- whether or not it's been updated/reviewed/etc (which feels like domain logic)

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
```diff
@@ -1,9 +1,32 @@
...
+- `label`: String; Name of Recipient
+- `status`: String; Response Status
```
mwilliamson-firefly, line 15: Does the status text always correspond with `uiState` (ignoring the selected state)? For instance, if resubmission if requested, we always show the same UI state and message? If so, perhaps it's worth just being able to pass a string like `resubmission-requested` in here to indicate the status, and then having the component decide on the message and styling?

blocks/core/ff_module/ff_module-profile-response-button/_ff_module-profile-response-button-renderer.js
```diff
@@ -10,9 +10,11 @@ module.exports = function() {
                 onSelect: function() {
                     console.log("onSelect");
                 },
-                isSelected: false,
+                uiState: "",
                 guid: "u42",
                 label: "Sally Student",
+                status: "Resubmission Requested",
+                mark: "B, 76%",
```
mwilliamson-firefly: In a branch, I have the mark being passed in a rich object i.e.

```
markAndGrade: {
    mark: 7,
    markMax: 10,
    grade: "A"
}
```

where some of the properties may be missing. That leaves the decision of how to actually format the mark and grade to the component, which seems to me like the right place to decide.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly: Ah, just saw it's used for the key. Would still be nice to know what `data-guid` is for though.

blocks/core/ff_module/ff_module-task-event-repeater/ff_module-task-event-repeater.rt
```diff
@@ -0,0 +1,11 @@
+<rt-require dependency="../../ff_module/ff_module-task-event/ff_module-task-event.js" as="TaskEvent"/>
+
+<div class="ff_container-task-event-repeater">
+    <ol class="ff_container-task-event-repeater__items">
+        <li class="ff_container-task-event-repeater__item"
+            key= "event-{eventIndex}"
```
mwilliamson-firefly: Is the intent to pass the list of events in the order that they should be rendered? The application currently stores them in chronological order (earliest first), but the designs I've seen shows them in reverse chronological order (latest first), which means that (assuming that this component renders in-order) the key for each event will change each time this component is rendered.

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
```diff
@@ -0,0 +1,6 @@
+<div
```
mwilliamson-firefly: I realised (somewhat belatedly) that "container" was a bad choice of name on my part since there's already an unrelated notion of container components in React. I struggled to think of an appropriate alternative name for `TaskEventContainer` though.

blocks/core/ff_module/ff_module-task-event/ff_module-task-event.js
```diff
@@ -0,0 +1,19 @@
...
+var StampResponseAsSeenTaskEvent = require('./src/_StampResponseAsSeenTaskEvent.js');
+var SetTaskEvent = require('./src/_SetTaskEvent.js');
+var AddedCommentEvent = require('./src/_AddedCommentEvent.js');
+
+var TaskEvent;
+
+var eventComponents = {
+    'set-task': SetTaskEvent,
+    'stamp-response-as-seen': StampResponseAsSeenTaskEvent,
+    'added-comment': AddedCommentEvent
```
mwilliamson-firefly: Given that the file with event types has been added to the pattern library, is it worth referencing them here rather using magic strings?

mwilliamson-firefly:
This seems sensible to me, just a few questions around where decisions around rendering should take place -- my (possibly incorrect) view being that we should pass rich objects around, and leave the decision of how to render to the leaf component.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: Will do.
Do you have a preference for convention? I see the React docs have return values at the start which seems a bit odd to me.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: The states aren't specific to melody, I'm just stating that the Melody css has implementations, but others won't yet. I can change the wording to make that clearer.
The uiState is just for managing css state modifiers. Perhaps different names?

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: I believe so. I don't think it's necessary for the React implementation but kept it in to remain consistent with the xsl implementation; can foresee other non-react scripts using that so thought it would be safest to keep it in.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly: My point is that managing CSS state modifiers is something that feels internal to the component. The props should describe the data that needs to be rendered, and the component is responsible for working what UI state that corresponds to. Considering just the data, whether or not this has been selected feels distinct from whether or not it's been (say) reviewed.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly: Okay. Personally, I'd leave it out until it's actually needed, but I'll leave it up to you.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods, line 15: No, the status is independent of the uiState. The status reflects the task status, the uiState only the state of the button (so its reviewed once the button has been clicked, updated if the status has changed but the button hasn't been clicked, selected is obvious, etc)

blocks/core/ff_module/ff_module-profile-response-button/_ff_module-profile-response-button-renderer.js
dkhuntrods: Ok, I'll update that.

blocks/core/ff_module/ff_module-task-event-repeater/ff_module-task-event-repeater.rt
dkhuntrods: I don't think I was considering order at all. Will the event have a key I can use, or what do you recommend to ensure order?

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly, line 15: How does the reviewed status behave between reloads? In other words, do we need to persist it?

blocks/core/ff_module/ff_module-task-event-repeater/ff_module-task-event-repeater.rt
mwilliamson-firefly: I think it's fine to assume that they're passed in chronological order, so this should just render them in reverse order. (Although I find the reverse order a bit confusing -- I assume others have already thought about this though?)

Each event has a local ID, so we could pass that through as a property on the event.

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
dkhuntrods: There are lots of containers in the patternlib though this doesn't entirely fit the convention. Perhaps a synonym like wrapper or something like base?

mwilliamson-firefly: Base feels a bit more appropriate than wrapper, so I'd be happy with that. (It's not really exposed anywhere outside of the pattern library though, so from Firefly's point of view it doesn't matter too much.)

blocks/core/ff_module/ff_module-task-event/ff_module-task-event.js
dkhuntrods: Ah, thought I had done. Will do.

dkhuntrods:
I agree that the leaf nodes should have responsibility - my concern would just be, how much might the definitions of the rich objects change, or are they pretty much fixed for now and in the future? If the latter, no problem.

mwilliamson-firefly:
They might change, but I'm not sure I see much of a problem with that?

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: I can't see the distinction between telling the component that its currently selected, and that it was previously selected? If you're passing in the fact that it's selected at all, that is?

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods, line 15: Yes, I think it needs to persist - it's a reflection of data having changed since the item was last viewed.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
mwilliamson-firefly: I thought "reviewed" corresponded to state based on the events (such as a response seen stamp), rather than just being UI state. Although if we're persisting the reviewed state, then I think it still requires a bit of thought (since selected is purely client-side, while the notion of reviewed is part of the wider data model, and therefore somewhat distinct).

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
dkhuntrods: Ok, it's pretty internal in the patternlib too, but not hard to change.

dkhuntrods:
It just means the component has to know more about the internal structure of the object, so the markup would potentially need to update to reflect that whenever the rich object def changes (if it's not a simple mapping) so we'd just have to ensure that changes were well communicated.

mwilliamson-firefly:
I think that's fine: if somebody's changing the structure in the Firefly code, then it should be clear that the component also needs updating (since the objects are built just for rendering components.) If not, tests should still catch it.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: Actually, the guid is used for the onSelect method in webapp

mwilliamson-firefly: Isn't it used by the list component rather than the button itself? (The point being that there's no need for a discriminator value when clicking a button since there's only thing that makes sense to have clicked.)

blocks/core/ff_module/ff_module-profile-response-button/_ff_module-profile-response-button-renderer.js
dkhuntrods: @mwilliamson-firefly  In the designs, there is a percentage displayed here. Should the component automatically convert mark/markMax, or is that a separate value that might be passed in?

mwilliamson-firefly: I'm not sure: I haven't seen anywhere that allows you to pick between a mark and percentage. We let the user pick in the markbook though. Perhaps one to ask Viv, I'll ping her on Trello. For now, let's just ignore percentages.

dkhuntrods: Alright!

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: Yes, it is. I'd thought the value was coming from the guid on the button but I see it's just a map of the props sent to the list.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: As far as I understand, it's distinct from the state of the events themselves. I had assumed that there was a direct correlation as well, but chatting to Rachel my understanding is that it just reflects if there's a new event in the event stream for that recipient since you looked at it. As I said, the names might not be ideal. It might not need to be persisted, that was just my instinct - my assumption is that it's like the read status on an email. Looking at the design and talking with Rachel, the ones that have all blue text are 'unread', the ones that have grey text have been 'read', the selected ones are blue background.

dkhuntrods: Perhaps we should double-check with Rachel?

mwilliamson-firefly: Hmm, I think regardless of whether it's persisted or not, it makes more sense to have separate `isSelected` and `isRead` properties, although it's definitely worth checking whether or not the read status should be persisted.

blocks/core/ff_module/ff_module-task-event-repeater/ff_module-task-event-repeater.rt
dkhuntrods: I assume so! 
Would the property be named `localEventId`?

mwilliamson-firefly: Good question. Let's go with `localEventId` for now, and we can change if it turns out that there's something more appropriate.

blocks/core/ff_module/ff_module-profile-response-button/ff_module-profile-response-button.md
dkhuntrods: Ok; I was attempting to avoid a whole mess of separate state vars, since from a display perspective they're mutually exclusive and I'd assumed it would only apply to ui state, but I can see that it's not necessarily just ui state here. I'll separate those out. Thanks!

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
```diff
@@ -0,0 +1,6 @@
...
+    class="ff_module-task-event ff_module-task-event--{this.props.event.type}"
+    >
+    <time class="ff_module-task-event__sent">{this.props.event.sent}</time>
```
mwilliamson-firefly: Would it make sense for this component to pick a date format, rather than expecting the date to already be formatted?

dkhuntrods: The dates can be as straight-up times, or as dates; in the xsl this had all been provided formatted so was assuming the same

dkhuntrods:
Ok! Revisions made and further comments welcome but hopefully this should do the trick! :) @mwilliamson-firefly

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
mwilliamson-firefly: Ah, okay, so it'll work fine with `Date` objects too?

mwilliamson-firefly:
Assuming the `sent` property works just fine when it's a `Date`, then I think this is good to merge.

blocks/core/ff_module/ff_module-task-event/src/TaskEventContainer.rt
dkhuntrods: Sorry, I meant in the design - sometimes it's just a time but sometimes it's a date (not a `Date` )
So it seemed like it was the same kind of fuzzy date logic as was being passed to the xsl templates, so I'm just rendering exactly what's provided...

mwilliamson-firefly: We'll need to implement whatever logic we want to use for fuzzy date logic here as well, and it feels to me like the best place to make that decision is in the component that's actually showing the date.

dkhuntrods: Alright, is that something that already lives in Firefly? 
If not, I'm guessing it's something that could end up being widely used... Dd you foresee a problem with it living in the patternlib?

mwilliamson-firefly: I don't know of anywhere in Firefly that already does it, but it's probably worth checking. Having it in the pattern library seems fine to me.

dkhuntrods:
I've updated this so the `sent` property will display with either Date or String objects.

PR #89, opened 2016-02-24
ff_container-expandable

sd-92:
Created the new container expandable and removed all the superseded patterns/code not used anymore


blocks/core/ff_container/ff_container-expandable/ff_container-expandable.js
```diff
@@ -0,0 +1,25 @@
...
+		
+		var $expandableOpenText = $expandableText.attr("data-open-text"),
+		$expandableClosedText = $expandableText.attr("data-closed-text");
+
+		var isVisible = $expandableContent.is(":visible");
+		var newText = isVisible ? $expandableClosedText : $expandableOpenText;
+
+		if (isVisible) {
+			$expandableContent.hide();
+			$expandableIcon.removeClass("ff_icon-page-up-open-blue").addClass("ff_icon-page-down-open-blue");
```
roobottom: Do we want to make this a bit more generic? I'm thinking of other themes that wouldn't use the blue icons. 

Hmm, but how to do this? Good question Jon, I'm glad I asked!

Well, first stage would be to pass the icons through in a data attr, maybe `data-ff-icon-closed="ff_icon-page-down-open-blue"`

Second, is how the theme could pass that through to the pattern. Have you any ideas on this @dkhuntrods / @mwilliamson-firefly ?

blocks/core/ff_container/ff_container-expandable/ff_container-expandable.xsl
```diff
@@ -0,0 +1,20 @@
+<xsl:template name="ff_container-expandable">
+	<xsl:param name="data" />
+	<div class="ff_container-expandable">
+		<div class="ff_container-expandable__dropdown ff_container-expandable__dropdown--small-screens" data-ff-action="expandable">
```
roobottom: Do you need the `class=""` in here if you're overwriting it with `xsl:attribute` ?

blocks/core/ff_container/ff_container-expandable/ff_container-expandable.js
mwilliamson-firefly: Why does the class need to specifically reference the colour? Can't you set it to something like `ff_icon-page-up-open` and let the theme pick the colour through CSS?

dkhuntrods: @mwilliamson-firefly @roobottom We should just be able to set a generic class, and then reference specific icons per theme. The problem is that the icons are generated from svgs using a build script, and the colours are generated too, as well as the corresponding css which inlines the svg. As it's a separate file, we can't really extend it in less. We can't really import into the theme file either as the page needs to figure out support and include the relevant file, or fallbacks if necessary. Will need to think more on this, as I've encountered it too.

dkhuntrods:
My only suggestion would be that a nice little animation might be nice here - try a css transition maybe, since it's just an enhancement, it's pretty well supported and won't effect performance as much as a js animation.

blocks/core/ff_container/ff_container-expandable/ff_container-expandable.xsl
sd-92: @roobottom Good point, fixed that!

blocks/core/ff_container/ff_container-expandable/ff_container-expandable.js
sd-92: @dkhuntrods @roobottom  does this mean that for now we will leave it as it is or should I change something?

sd-92:
@dkhuntrods what kind of animation were you thinking about? like slide or fade? I think if it should animate then sliding will look better.

blocks/core/ff_container/ff_container-expandable/ff_container-expandable.js
roobottom: @Simonedeluca I still think it should be more generic anyway when we swap out icons in js. The icon should come from a data attr, not be embedded in the js.

sd-92: @roobottom agreed! I've updated that so that it gets everything from the data attribute passed to the pattern.

dkhuntrods:
@simonedeluca-firefly I'd agree, a sliding animation would be best I think; pretty quick with an ease out would be my first thought :)

PR #90, opened 2016-02-24
Overlay bar is now overlay

roobottom:
I've updated the overlay-bar container to be more generic. 

See inline comments for more info.


blocks/core/ff_module/ff_module-button/_ff_module-button-renderer.js
```diff
@@ -6,7 +6,7 @@ var Button = require('./ff_module-button');
 var buttonProps = [{
     text: 'Button default',
     id: 'btn-01'
-}{
+},{
```
roobottom: This was causing a build error. Assume its fixed elsewhere, but just incase, it's here too.

src/gulp-plugins/gulp-mdtoXSLT.js
```diff
@@ -42,7 +42,7 @@ module.exports = function mdtoXSLT(options) {
     return through.obj(function(file, enc, callback) {
         var context = getContext(file);
         var t = this;
-        output(context);
+        // output(context);
```
roobottom: These were not entirely useful. Happy to back this out if you think we should leave them in.

roobottom:
@dkhuntrods I've updated this PR. Can you have a look see when you get time please? Ta.

src/gulp-plugins/gulp-mdtoXSLT.js
dkhuntrods: Can you add the debug: false param back in to lib/mdToXslt{Blocks,Pages}.js rather than removing the code here? That should stop the debug info rendering with every update, but allows us to view it when necessary.
EDIT: Or did you mean the overlays.md file? If it wasn't useful in the end please do remove yeah.

PR #93, opened 2016-03-01
Feature task details response

dkhuntrods:
Apologies, I've included the list of test pages that I'd added back in this branch. 
Those are the old ones, so perhaps don't worry about them for now.
@mwilliamson-firefly if you could have a look when you get a chance that would be lovely.


blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component.md
```diff
@@ -0,0 +1,17 @@
...
+# Props 
+- **text**: {String, required} The text rendered by the button
+- **id**: {String, required} The -target and -trigger id used by the core dropdown js file.
+- **modifier**: {String}: Add additional css styling for different effects.
+- **list**: {Array, required} An Array of Objects with { text: {String}, href: {String} } properties
+- **isDisabled**: {Boolean} Whether or not the dropdown can be clicked.
+- **isOpen**: {Boolean} If true, the list will be rendered in the open state.
+
+
+**Ensure the core dropdown file is included on a page, and activated on DOM ready or equivalent, to activate js functionality.**
```
mwilliamson-firefly: This seems like strange behaviour. Why can't the React component activate itself?

dkhuntrods: The functionality was already contained in a pre-existing module. I could allow the React component to call the activation method, but then it would need a separate set of data-attributes, so that if the vanilla method was called on the page it wouldn't clash. ~~I thought having a consistent set of data-attributes was more important, and the likelihood high that the vanilla activation method would already be called on a page.~~ Upon further consideration, I agree. I'll look at the best way to have this activate automatically.

blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = React.createClass({
+    displayName:'DropdownButton',
+    propTypes: {
+        list: React.PropTypes.array.isRequired,
+        id: React.PropTypes.string.isRequired,
+        text: React.PropTypes.string.isRequired
+    },
+    render: template,
+    componentDidMount: function(){
+        activateDropdowns();
```
mwilliamson-firefly: It feels to me as though the least surprising behaviour would be to activate dropdowns just for this component to avoid weird interaction between React and non-React usages of the dropdown. Since `activateDropdowns` lets you both pick the root and the selectors, it seems like that wouldn't be too tricky to implement?

blocks/core/ff_module/ff_module-button/ff_module-button.rt
```diff
@@ -1,15 +1,18 @@
-<button type="button"
+<button
+    type="button"
     title="{this.props.text}"
     id="{this.props.id}"
     disabled="{this.props.disabled}"
-    class="ff_module-button{this.props.modifier ? ' ff_module-button--'+this.props.modifier : ''}{this.props.classes ? ' '+this.props.classes : ''}{this.props.disabled ? ' ff_module-button--is-disabled' : ''}"
+    class="{this.generateClass('ff_module-button', this.props)}"
```
mwilliamson-firefly: Do you need to pass `this.props`? Presumably `generateClass()` already has access to them? Similarly, is there ever an occasion where it's called without `'ff_module-button'` as the first argument?

blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component.md
```diff
@@ -0,0 +1,18 @@
...
+
+See [ff-module-dropdown-button](/blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button.html) for details on dropdown styling and modifiers.
+
+This component just renders an html dropdown button, it doesn't replicate the dropdown functionality which already exists in the core 'vanilla' dropdown file 'ff_module-dropdown.js'.
+
+**The core dropdown file is activated automatically once the component mounts.**
+
+# Props 
+- **text**: {String, required} The text rendered by the button
+- **id**: {String, required} The -target and -trigger id used by the core dropdown js file.
```
mwilliamson-firefly: I'm not sure I understand what this is used for. Is this just an ID that needs to be unique on the page for use just within this component? If so, couldn't you just automatically generate it in the component?

blocks/core/ff_module/ff_module-task-event/_src/events.js
```diff
@@ -20,6 +14,6 @@ module.exports.types = {
 var nextEventId = 1;
 // Local event IDs are local to the client
 // They have no meaning except to uniquely identify events
-exports.generateLocalEventId = function generateLocalEventId() {
+module.exports.generateLocalEventId = function generateLocalEventId() {
```
mwilliamson-firefly: Does `generateLocalEventId()` get used by the pattern library anywhere?

mwilliamson-firefly:
I was focusing on the actual implementations of the React components rather than other bits like tests and CSS. Generally looks fine, left a couple of thoughts.

blocks/core/ff_module/ff_module-task-event/_src/events.js
dkhuntrods: No, I thought I'd taken that out actually.

blocks/core/ff_module/ff_module-button/ff_module-button.rt
dkhuntrods: Not strictly, no, for the props. I've kept the base separate just for consistency as that's a common pattern in the components and often it does change within a component to apply to different elements.

blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component.js
dkhuntrods: I had thought about that but was ambivalent on which approach would be better. If you've got a strong sense of it, happy to do so.

mwilliamson-firefly: Yeah, I feel like keeping the behaviour separate will cause less strange interaction between components if you're happy with it that way.

blocks/core/ff_module/ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component.md
dkhuntrods: It is, but it's also used to generate an actual `id` on the element, so needed to be externally defined. I could change the default behaviour to use the `id` if present, but generate an internal trigger/target id if not.

PR #94, opened 2016-03-01
Updates to the expandable container

sd-92:


blocks/melody/ff_container/ff_container-expandable/ff_container-expandable.less
```diff
@@ -1,29 +1,36 @@
...
 	}
-}
 
-.ff_container-expandable__icon  {
-	width: 1.5em;
-	height: 1.5em;
-	float: right;
-}
+	&__content {
+		max-height: 1000px;
```
roobottom, line 17: I don't really understand why you're setting a height of `1000px` here? I thought the idea was to have a modifier like `--fixed-height` that would set a max-height of say, 200px, and an `overflow-y: auto` ? This seems like it would hide everything after a 1000px.

blocks/melody/ff_container/ff_container-expandable/ff_container-expandable.less
```diff
@@ -1,29 +1,36 @@
...
-	max-height: 0;
-}
+	&--is-widescreen &__header {
+		@media all and (min-width: @ff_bp_wide) { 
+			display: none;
+		}
+	}
+	&--is-widescreen &__content {
+		@media all and (min-width: @ff_bp_wide) { 
+			max-height: 1000px;
```
roobottom, line 33: Also, why is this being set here again?

blocks/melody/ff_container/ff_container-expandable/ff_container-expandable.less
sd-92, line 17: I am not sure we are talking about the same thing. This is not the filters, that was merged by Danielle and gives the option to have a max-height through a modifier (as we discussed) and then a scroll bar to scroll the filters. The expandable component needs a high max-height to do the sliding animation (so we set it to a high value in this case 1000px). Should we have a higher value for the max-height @dkhuntrods ?

blocks/melody/ff_container/ff_container-expandable/ff_container-expandable.less
sd-92, line 33: When It's `--is-widescreen` (which means content should show and heading should disappear on widescreen) the javascript will apply the class of `is-collapsed` regardless of it.  `is-collapsed` sets the max-height: 0 to hide the content. So I want to set the max-height of the content to make it visible on widescreens. This max-height thing is only for animation purposes.

blocks/melody/ff_container/ff_container-expandable/ff_container-expandable.less
roobottom, line 17: OK, well I'm unclear then on why we need to set a max height in order for the animation to work — this feels a bit hacky. Can you explain why this is?

sd-92, line 17: From my understanding, in order to animate the height of an element in CSS (to create a slide-up/slide-down) you must set either a height or a max-height on the element. It can't animate from 0 to auto(this is true for other properties as well). This could have been done in JQuery with a slideDown/SlideUp but we decided to use a CSS animation which would be less expensive. I am not an expert when it comes to CSS animations but I think this is the reason.

roobottom, line 17: Hmm, this _does_ seem like the accepted way of doing things (after some googling around).

PR #95, opened 2016-03-03
Update of overlay

roobottom:
- changed padding -> margin. 
- Increased spacing. 
- Removed modifier from lib_test version of overlay


dkhuntrods:
Let's just keep an eye on the change to margin spacing & look out for collapsing margins in implementation.

PR #96, opened 2016-03-04
Made the control bar container responsive

roobottom:


blocks/melody/ff_module/ff_module-button/ff_module-button.less
```diff
@@ -6,6 +6,7 @@
         border: 1px solid @ff_colour_default;
         padding: (@ff_size_spacing_min * .85) @ff_size_spacing_med;
         cursor: pointer;
+        white-space: nowrap;
```
roobottom: This is to ensure that text on the button remains on one line, despite pressure from any containing elements. @aegirfirefly can you please confirm this is as designed?

dkhuntrods:
So to recap, the groups should behave the same whatever number of groups there are, and the responsive modifiers should reflect their full-size counterpart (so a -right should still be right-aligned in responsive, where it makes sense)

PR #107, opened 2016-03-11
Adding some initial coding guidelines for React based patterns

AdamLynamFirefly:


README.md
```diff
@@ -90,6 +90,32 @@ If the core file requires other services or controls, we can mock them with an u
...
+Separate template files are required if any of the following conditions are true for a given pattern:
+
+ - The pattern is JavaScript/React only. The absense of an XSL means a template is required to quickly see the shape of a patterns output.
+ - The pattern outputs more than one DOM element. As soon as the complexity of a pattern increases to contain more than one HTML DOM element a template is required to quickly see the shape of a patterns output.
+
+In most cases these templates should naturally grow from raw HTML files used to plan the pattern.
+
+**JSX Files**
+
+Moving dom rendering into a `.jsx` file is a convenient way to separate DOM structure from pattern set up. Id's, classes and data attributes being set on DOM elements should be as clear as is practical to be (try not to hide attribute values inside props if it can be avoided in the pattern usage).
```
dkhuntrods, line 117: Could you clarify what you mean by 'hiding attribute values inside props'?

dkhuntrods:
I think that's great Adam, thanks very much for writing this up. Just a couple comments/questions; see inline.

README.md
```diff
@@ -90,6 +90,32 @@ If the core file requires other services or controls, we can mock them with an u
...
+##### JSX
+
+JSX syntax is available, and should be used whenever practical to replace calls to React.createElement() due to the transparency of the DOM structure it offers.
+
+##### Separate Template Files
+
+Separate template files are required if any of the following conditions are true for a given pattern:
+
+ - The pattern is JavaScript/React only. The absense of an XSL means a template is required to quickly see the shape of a patterns output.
+ - The pattern outputs more than one DOM element. As soon as the complexity of a pattern increases to contain more than one HTML DOM element a template is required to quickly see the shape of a patterns output.
```
dkhuntrods: I'm just wondering if this is the most useful condition we can make for the requirement of a template file.

README.md
AdamLynamFirefly, line 117: I was thinking, we probably want to avoid writing code like:

`<a className="this.props.className" />`

And prefer to put as much static information into the template files as possible, like this:

`<a className="ff_module-my-module ff_module-my-module--{this.props.modifier}" />`

Maybe I should put that example in the README explicitly. If you agree we should try and do that of course (I know it won't always be practical).

README.md
AdamLynamFirefly: I briefly discussed this with @jon, but I admit I don't have a firm grasp of what we are aiming to do with templates. I suspect that is because of my function-over-form driven approach to front-end that I just don't understand the pressure to include them.

I do think it is valuable to have some guidelines and expectations though, so I wanted to get something written down as a talking point. It would be great to discuss what we are aiming to achieve here.

README.md
dkhuntrods, line 117: Oh definitely. For a lot of the patterns I've written a helper generateClass class, with the 'base' (ie. 'ff_module-my-module') specified in the 'template' - makes it a bit cleaner to combine the 'classes' prop as well, but still keeps the static information in the 'template'. What do you think of that?

README.md
dkhuntrods: I agree. I think this may be a case of, let's build some and see what arises, rather than imposing without full knowledge, as has happened once or twice already :P.
What do you say @roobottom?

README.md
AdamLynamFirefly, line 117: That sounds ideal. Where can I find an example of your generateClass helper. I think that would be perfect to explicitly reference here.

dkhuntrods, line 117: I think the button module is a pretty good example, as there are of states, classes, modifiers, and combinations of the above in there.

README.md
roobottom: I think the need is captured better in your next point @AdamLynamFirefly 

> In most cases these templates should naturally grow from raw HTML files used to plan the pattern.

I think the templates that are sufficiently simple enough to build directly into React, then no problem. Those that are more complex then how about we set a rule that either:
- There must be a flat HTML or XSLT version built first
- There must be some documentation included in the main `.md` file that contains the raw HTML for the component, and sufficient explanation of states and any classes that get added. 

The only thing I'm trying to cover off here is the writing / reviewing of complex markup, and getting that right before including that in the React.

dkhuntrods: I've not been finding separate templates necessary - when there's any complexity, it doesn't seem to help but rather obscure things. 

My suggestion at this time would be: If a component involves rendering decisions or complex dynamic aspects, you _may_ separate things into template files where it makes sense to do so, and _do_ include the flat HTML in the .md file.

PR #109, opened 2016-03-14
ff_container-filter react

sd-92:


dkhuntrods:
So the original functionality this is replacing, is gone now? This does nothing but display a list of child elements, am I correct? Just checking I'm not missing something

PR #111, opened 2016-03-16
Dkh feature task response

dkhuntrods:
Adding the updated Task Responses screen, with an accessor to it's rendered Modal component, and a stub for the Modal component that currently just references the default npm 'react-modal' package we're using.


sd-92:
LGTM

PR #112, opened 2016-03-16
ff_container-expandable-component react

sd-92:


blocks/core/ff_container/ff_container-expandable/ff_container-expandable-component/_ff_container-expandable-component-renderer.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+var React = require('react');
+
+var Expandable = require('./ff_container-expandable-component');
+
+var data = {
+	expdanded_text: 'Hide content',
```
dkhuntrods: Hi Simone, purely since we're trying to be consistent across projects, can you use camelCasing for js vars? Thanks!

blocks/core/ff_container/ff_container-expandable/ff_container-expandable-component/ff_container-expandable-component.js
```diff
@@ -0,0 +1,25 @@
...
+			</span>
+			<span data-icon='expandable-icon' data-collapsed-icon={this.props.collapsed_icon} data-expanded-icon={this.props.expanded_icon} className={'ff_icon ff_container-expandable__icon ' + this.props.collapsed_icon}></span>
+		</div>
+		<div className='ff_container-expandable__content' data-ff='expandable-content'>
+			{this.props.content}
+		</div>
+		</div>;	
+	},
+	componentDidMount: function() {
+		Expandable();
```
dkhuntrods: As this is now, the Expandable method will be called for every instance of this component on the page, and so each instance will get multiple handlers which could start to cancel each other out or behave strangely. I'd recommend adding an options object, similarly to the dropdown, where the root node can be specified so that all the expandable actions only take place in the context of the rendered react expandable node.

PR #115, opened 2016-03-18
Module progress react

roobottom:
A React version of the progress bar.

### DIFF
--- a/tests/ff_module-progress.test.js
+++ b/tests/ff_module-progress.test.js
@@ -1,0 +1,28 @@
+'use strict';
+
+var React = require('react');
+require('./lib/utils').bootstrapBrowser();
+var TestUtils = require('react-addons-test-utils');
+var expect = require('chai').expect;
+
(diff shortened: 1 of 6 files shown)


tests/ff_module-progress.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+describe('Progress', function() {
+
+  it('should always render a progress bar', function() {
+    var element = React.createElement(Progress, {})
+		var component = TestUtils.renderIntoDocument(element);
+		expect(component).to.exist;
+  })
+
+  it('should always render a progress bar from minimal data', function() {
```
dkhuntrods, line 18: You should probably also test that the bar is rendering expected values from provided data, like x marked is x% wide (you could add a set-width wrapper if the width is percentage based)

roobottom, line 18: I added some more robust tests @dkhuntrods — can you take another look. Ta. (The code isn't very DRY, I didn't know if this was a bad thing in test files like this?)

PR #116, opened 2016-03-21
Dkh task response details dropdown

dkhuntrods:
@roobottom or @simonedeluca-firefly If you could take a quick look: small updates & little fixes


sd-92:
Looking good! :dart:

PR #121, opened 2016-03-30
Add .idea/ to .gitignore

robbie-c:
This is used by JetBrains IDEs such as WebStorm.

Should be a harmless change but useful for those who use it, as otherwise their git change list will be polluted.


AdamLynamFirefly:
LGTM

PR #124, opened 2016-04-06
Dkh modal dialogs

dkhuntrods:


blocks/core/ff_container/ff_container-dialog/ff_container-dialog.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+var React = require('react');
+
+function generateClass(base, props){
+    var classNames = [];
+        classNames.push(base);
+        if (!!props.modifier) classNames.push(base + '--' + props.modifier);
```
robbie-c, line 8: Tiny: Do you need the `!!` here?

dkhuntrods, line 8: I can't remember the exact case, but there was an instance where the distinction between truthy and actual boolean value was required, so this is just to make that explicit.

blocks/core/ff_module/ff_module-form-add-attachment/ff_module-form-add-attachment.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+var React = require('react'),
+    DropdownButton = require('../ff_module-dropdown-button/ff_module-dropdown-button-component/ff_module-dropdown-button-component');
+
+function renderFileList(props) {
+    var option = null;
+    switch (true) {
```
robbie-c: This looks like it should be an `if` statement.

blocks/core/ff_module/ff_module-form-add-attachment/ff_module-form-add-attachment.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+    onDragLeave: function(){
+        this.setState({
+            dragCounter: decrementCounter(this.state.dragCounter)
+        });
+    },
+    onDragOver: function(e) {
+        e.preventDefault();
+        e.stopPropagation();
+        return false;
```
robbie-c: I don't think you need to return false here, I think calling `stopPropagation` should handle that.

blocks/core/ff_container/ff_container-dialog/ff_container-dialog.js
robbie-c, line 8: I think the distinction should only matter when testing equality. For example later on you've used `switch(true)` which is going to test equality with `true` rather than testing truthiness, but an `if` will test truthiness so you should be alright without it.

blocks/core/ff_module/ff_module-form-add-attachment/ff_module-form-add-attachment.js
dkhuntrods: Ok, had grabbed that from the react-dropzone component; I believe it's for older msie browsers but I suppose if we're not worrying about them you're absolutely right.

blocks/core/ff_module/ff_module-form-add-attachment/ff_module-form-add-attachment.js
dkhuntrods: Ok

blocks/core/ff_module/ff_module-form-input/ff_module-form-input.js
```diff
@@ -2,10 +2,23 @@
...
+    return classNames.join(' ');
+}
+
 module.exports = React.createClass({
 	displayName: 'FormInput',
 	render: function() {
-		var attributes = {};
+
+		var attributes = {},
+            className = generateClass('ff_module-form-input', this.props);
```
robbie-c, line 20: Tiny: This might just be github's diff view, but the indentation is weird on this line. I think it might be spaces where everything else is tabs.

It might be worth using something like http://editorconfig.org/ to keep this consistent even across all the various editors that people use, but that's beyond the scope of this PR :P

blocks/core/lib_test/lib_test-task-responses-screen/_lib_test-task-responses-screen-renderer.js
```diff
@@ -184,22 +186,29 @@ var overlayInner = React.createElement(TaskResponses, {events: events,
...
     document.addEventListener('DOMContentLoaded', function(event) {
         var el = document.querySelector('[data-lib_test-task-responses-screen]'); //Use jquery or sim in Firefly for backwards compat
         if (el) {
 
-            var element =
-            <div className="ff_module-task-responses">
+            var element = <div className="ff_module-task-responses">
                 <div className="ff_util-row-bottom">
-                    <ContainerControlBar modifier= "split">
+                     <ContainerControlBar modifier="split">
```
robbie-c: Tiny: space here messes up indentation

blocks/core/ff_module/ff_module-form-add-attachment/ff_module-form-add-attachment.js
```diff
@@ -4,13 +4,10 @@ var React = require('react'),
...
-    switch (true) {
-        case (props.files && props.files.length>0):
-            option = renderWithFiles(props);
-            break;
-        default:
-            option = renderNoFiles();
+    var option = renderNoFiles();
+
+    if (props.files && props.files.length > 0) {
+        option = renderWithFiles(props);
```
robbie-c, line 10: I think you could probably simplify this further to

``` javascript
if (props.files && props.files.length > 0) {
    return renderWithFiles(props);
} else {
    return renderNoFiles();
}
```

blocks/core/ff_container/ff_container-dialog/ff_container-dialog.js
dkhuntrods, line 8: I know the distinction there, I'm just saying that there were some actual values values for modifier and classes that produced false positives without the doublebang, though as I say I can't recall the exact one.

robbie-c:
Other than a few small changes, the React code LGTM :)

PR #125, opened 2016-04-08
created ff_module-columnar-list pattern

boonier:
Created a new pattern module for responsive columnar lists.


roobottom:
Worked on this alongside Si. LGTM.

PR #126, opened 2016-04-11
Dkh task event variations

dkhuntrods:
Some refactoring (remove unnecessary file underscores, replace .rt files with inline .jsx); then 
- adding file React component
- adding file-list React component
- adding task events: 
  - add-file
  - delete-response
  - confirm-student-is-unexcused


blocks/core/ff_module/ff_module-task-event/_src/MarkAndGradeTaskEvent.js
```diff
@@ -0,0 +1,64 @@
...
+    switch (true) {
+        case !!(event.mark && !event.grade):
+            return base + 'mark';
+            break;
+        case !!(!event.mark && event.grade):
+            return base + 'grade';
+            break;
+        case !!(event.mark && event.grade):
+            return base + 'mark and grade';
+            break;
```
robbie-c: I don't think a switch statement is the best way to handle this. I feel like an `if else` would be much clearer:

``` javascript
if (event.mark && !event.grade) {
  return base + 'mark';
} else if (!event.mark && event.grade) {
  return base + 'grade';
} else if (event.mark && event.grade) {
  return base + 'mark and grade';
}
```

Is there a reason that I haven't considered that a `switch` statement might be better? I think:
- `if` is shorter
- `switch(true)` needs the `!!` which is easy to forget
- `if` is a lot more idiomatic. This is a bit fluffy as a reason but I think probably the most important. `switch` statements are not really intended to have a constant expression. Whereas this function is a textbox example of where to use `if else`

dkhuntrods: Sure, I think I just have a slight anti `if-else` bias that tends me towards switch for more than three cases. In this case I agree, `if-else` would probably be more appropriate.
Aside from that, does it look ok?

robbie-c:
LGTM

blocks/core/ff_module/ff_module-task-event/_src/MarkAndGradeTaskEvent.js
robinmessage: FWIW, maybe nesting the `if`s would help?

> ```
> ```

if (event.mark) {

  if (event.grade) {

```
return base + 'mark and grade';
```

  } else {

```
return base + 'mark';
```

  }

}  else if (event.grade) {
  return base + 'grade';
}

```



-- 


[image: Firefly] <http://fireflysolutions.co.uk/>
Robin Message - Head of Technology
Firefly Learning LLP, 1 Lyric Square, London W6 0NB, UK

Support: +44 (0)20 7112 9362 <+44+(0)20+7112+9362> | Sales: +44 (0)20 8133
4415 <+44+(0)20+8133+4415> | fireflylearning.com | @fireflyteam
<http://twitter.com/fireflyteam>
Registered in England and Wales with partnership number OC336950.
VAT number: 981 8080 93.
```

PR #128, opened 2016-04-12
Added new container type: ff_container-landing-section

boonier:
Container acts as a wrapper around content, for example ff_module-columnar-lists.


blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.less
```diff
@@ -1,38 +0,0 @@
-//header
```
dkhuntrods, line 1: The core file should have some basic styling

blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.xsl
```diff
@@ -1,15 +1,17 @@
...
+        <header class="ff_container-dashboard-section__header ff_container-dashboard-section__header--{$data/item/@modifiers}">
             <h2 class="ff_container-dashboard-section__title"><xsl:value-of select="$data/item/@title"/></h2>
         </header>
         <div>
             <xsl:attribute name="class">
                 <xsl:text>ff_container-dashboard-section__main</xsl:text>
                 <xsl:if test="$data/item/@show-footer = 'no'"> ff_container-dashboard-section__main--no-footer</xsl:if>
             </xsl:attribute>
-            <p>Modules Go here</p>
+            <xsl:for-each select="$data//module">
```
dkhuntrods: I think it makes more sense for this to be a single node named something like `content` rather than a list of `module`s as it'll always be a single node being passed in to the container.

blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.xml
```diff
@@ -1 +1,7 @@
-<item title="{{title | default('Container Title')}}" more-page-text="{{pagetext}}" more-page-link="{{ pagelink | default('#') }}" show-footer="{{ showfooter | default('yes') }}" modifiers="{{ modifiers }}"/>
+<item title="{{title | default('Container Title')}}" more-page-text="{{pagetext}}" more-page-link="{{ pagelink | default('#') }}" show-footer="{{ showfooter | default('yes') }}" modifiers="{{ modifiers }}">
+	{% if modules %}
```
dkhuntrods: I think it makes more sense for this to be a single node named something like `content` rather than a list of `module`s as it'll always be a single node being passed in to the container.

blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.less
roobottom, line 1: @dkhuntrods I'm not sure that's the case here — I don't even think this file should exist. The only core styles _should_ be things like layout — as this doesn't have layout that's shared between themes, it doesn't need core styling. Unless I'm missing something?

dkhuntrods, line 1: @roobottom Fair enough! Yeah looking more closely I do see that it's all pretty much theme specific stuff so that makes sense. In which case, yeah file should probably go.

boonier, line 1: @dkhuntrods this is deleted now

blocks/core/ff_container/ff_container-dashboard-section/ff_container-dashboard-section.xsl
boonier: @dkhuntrods have updated this now

dkhuntrods:
LGTM

PR #130, opened 2016-04-13
Sd form preview updates

sd-92:


blocks/core/ff_module/ff_module-form-preview/_ff_module-form-preview-renderer.js
```diff
@@ -2,32 +2,102 @@
...
+			sent_to: 23,
+			no_excused: 2,
+			completed_by: 20,
+			marked: 3
+		},
+	},
+	{	
+		modifier: 'controls',
+		title: 'Marking Controls', 
+		markingControls: true
```
dkhuntrods: You can just pass the MarkingControls object here, rather than having it as a separate prop. Then it could be made more generic , so different components could be passed in - so instead of 'markingControls' the prop name could be 'component' .

blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.js
```diff
@@ -17,32 +18,52 @@ module.exports = React.createClass({
...
-							{item.list.map((item)=>{
-								return [<dt className='ff_module-form-preview__sublist__title'>{item.title}</dt>,
-								<dd className='ff_module-form-preview__sublist__data' data-ff-preview-for={item.previewfor}>{item.value}</dd>]
-							})}
-						</dl>;
-					} else if(item.html) {
-						listItemData = <div className='ff_module-form-preview__list__description'>{item.html}</div>
-					} else {
-						listItemData = item.value;
+					switch(true) {
```
dkhuntrods: Probably best to use `if/else if` here rather than `switch`

blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.js
```diff
@@ -17,32 +18,52 @@ module.exports = React.createClass({
...
+								})}
+							</dl>;
+							break;
+						case !!item.html:
+							itemData = <div className='ff_module-form-preview__list__description'>{item.html}</div>;
+							break;
+						case !!item.progress:
+							itemData = <div className="ff_module-form-preview__progress"><Progress {...item.progress}/></div>;
+							break;
+						case !!item.markingControls:
```
dkhuntrods: So with my other comment, this would be

```
if (item.component) {
    itemData = <div etc>{item.component}</div>
}
```

blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.js
```diff
@@ -11,38 +12,55 @@ module.exports = React.createClass({
 				title: React.PropTypes.string.isRequired,
 				url: React.PropTypes.string,
 				value: React.PropTypes.string,
+				value: React.PropTypes.modifier,
```
dkhuntrods: Should these both be `value`?

dkhuntrods:
Once that propType conflict is fixed, look good to me

PR #141, opened 2016-04-19
Update RecipientPicker to allow for autofocus to be turned off

dkhuntrods:
Turn off the autofocus on first page load - subsequent adding/selecting of items will still re-focus on input field. By default autofocus is still on, to maintain current behaviour.


roobottom:
@dkhuntrods is the intention that we'll set a variable that turns this behavior on in the pattern lib, but that wouldn't get exported?

dkhuntrods:
@roobottom The variable would be set only in the renderer, yes. The default behaviour wouldn't change.

PR #146, opened 2016-04-22
Wrap event and Actions in an object

bwfox:


bwfox:
I'm sure some additions are needed for this, such as styling, but hopefully it's a useful starting point.

PR #147, opened 2016-04-25
Si ff container landing section

boonier:
- Inject svg directly into the dom for dynamic styling
- Add Aegir's .colour-guard() mixin  to handle contrast colours
- Create .link() mixin to handle contrast colours for landing section links
- Update .title() mixin to handle contrast colours for landing section titles
- Remove grunticon.svgLoadedCallback from head.xsl as not required yet

### DIFF
--- a/blocks/melody/_shared/mixins.less
+++ b/blocks/melody/_shared/mixins.less
@@ -114,9 +114,6 @@
     }
 }
 
-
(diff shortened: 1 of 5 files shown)


blocks/melody/_shared/mixins.less
```diff
@@ -114,11 +114,37 @@
     }
 }
 
-
+// Calculating contrasts from here: http://24ways.org/2010/calculating-color-contrast/
+.colour-guard(@colour, @dark: #000000, @light: #ffffff) when ((((red(@colour)*299)+(green(@colour)*587)+(blue(@colour)*114))/1000) > 204) {
```
dkhuntrods, line 118: These probably shouldn't be in a non-prefixed context; they're only useful for text for so could go somewhere in prose utils.

blocks/melody/_shared/mixins.less
```diff
@@ -114,11 +114,37 @@
...
+    color: @light;
+}
+.colour-guard-svg(@colour, @dark: #000000, @light: #ffffff) when ((((red(@colour)*299)+(green(@colour)*587)+(blue(@colour)*114))/1000) > 204) {
+    fill: @dark;
+}
+.colour-guard-svg(@colour, @dark: #000000, @light: #ffffff) when ((((red(@colour)*299)+(green(@colour)*587)+(blue(@colour)*114))/1000) =< 204) {
+    fill: @light;
+}
 
 #ff_mix-sections {
```
dkhuntrods: Is this used anywhere aside from the container-landing sections? If not it should probably just live inside the container-landing less file.

dkhuntrods, line 131: Is this used anywhere aside from the container-landing sections? If not it should probably just live inside the container-landing less file.

dkhuntrods:
@boonier Couple small comments, otherwise look good.

PR #148, opened 2016-04-25
Dkh wrap events

dkhuntrods:
@bwfox I've updated the main files involved; I think the Modals, Editors and RecipientLists will need to be updated eventually as they've got some reflection of state too, but we can cross that bridge as we come to it...
If you could take a look and make sure it fits with what you're doing that'd be great.


bwfox:
LGTM

PR #153, opened 2016-04-26
Overview tabs to react

sd-92:


blocks/core/ff_module/ff_module-tabs-navigation/ff_module-tabs-navigation.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+var React = require('react');
+var TabsControl = require('../ff_module-tabs/_ff_module-tabs-renderer.js');
```
dkhuntrods: Renderers aren't ever meant to be exported or used directly

blocks/core/ff_module/ff_module-tabs-component/ff_module-tabs-component.js
```diff
@@ -0,0 +1,37 @@
...
+				state: React.PropTypes.string,
+				label: React.PropTypes.string.isRequired,
+				content: React.PropTypes.element.isRequired,
+				id: React.PropTypes.string.isRequired,
+				key: React.PropTypes.number.isRequired
+			})
+		).isRequired
+	},
+	componentDidMount: function() {
+		var options = {
```
dkhuntrods: I'd add distinct target and trigger data-attributes here, to keep the functionality of the React component really separate from the vanilla one. Let me know (or have a look at dropdown-button-component) if you're not sure what I mean :) For the tabs, the current values are : 
`linkSelBase: 'data-ff-tabs-target',
 contentSelBase: 'data-ff-tabs-content',`

tests/ff_module-tabs-component.test.js
```diff
@@ -0,0 +1,56 @@
...
+	});
+
+	it('should render tabs with a modifier: ' + modifier, function() {
+		var element = React.createElement(TabsComponent, {modifier: modifier, tabs: tabs});
+		var component = TestUtils.renderIntoDocument(element);
+		var root = TestUtils.scryRenderedDOMComponentsWithClass(component, 'ff_container-tabs-content');
+		expect(root[0].className).to.contain('ff_container-tabs-content ff_container-tabs-content--' + modifier);
+	});
+
+	it('should activate the clicked tab', function() {
```
dkhuntrods: I think you should also check that the correct tab and navigation are both selected after a click

dkhuntrods:
Just wondering, was it always the intention to reuse the existing tabs code, or did we want to re-do this with better validation handling, etc, when we rebuilt it in React?

sd-92:
@dkhuntrods I thought we wanted to use what we already had. But if you think that it could be made better in the react version, then I think this could be a good chance to do it.

dkhuntrods:
@Simonedeluca Ok, well we need the React tabs for the overview screen, and we need to implement the React validation on that screen, so I think we might need to make this a pure React solution. Can you double-check with Robbie (that's who's helping wire up Overview, yeah?)

robbie-c:
@dkhuntrods What do you mean by "pure React solution"? Do you mean the navigation events should go through our JS rather than through the DOM (e.g. passing in props `onTabSelect()` and `selectedTabIndex` rather than using `<a href="#id-of-the-thing-to-select">`)?

dkhuntrods:
@robbie-c Yeah, rather than wrapping the existing tab handling, we'll want the React Tabs to have its own validation through props, so it can integrate with the proposed React validation. Unless you can see a straightforward way for the wrapped tabs to integrate with a Validation component?

robbie-c:
Possibly stupid question: what do you mean by "Validation" in this context? How does a Validation component work? Is there an example I can look at the source of?

dkhuntrods:
It doesn't yet exist - what we all discussed in the validation meeting was using an external react validation module. For forms, etc; though one in particular could be used for arbitrary components.
Like this: https://github.com/christianalfoni/formsy-react

sd-92:
@dkhuntrods I'll give it a go using that react validation module!

dkhuntrods:
@simonedeluca-firefly @robbie-c  Sorry, Simone, we haven't decided on a Validation component yet; that was just an example. The tabs version just probably needs to deal with its own state, rather than relying on the existing tabs functionality, so that it _can_ use other react validation components down the road, but we're still discussing how necessary this is.

robbie-c:
OK so my problem with the implementation as it stands is that it manages its own state (in this case, which tab is selected). I think we'd like as much state as possible in the application layer, and have the selected tab passed in as a prop. The tabs also need a way of signalling when a new tab is selected, so it can take a function in as a prop.

robbie-c:
So for the tabs nagivation, you might haves something like

``` javascript
propTypes: {
    onTabSelected: React.PropTypes.function.isRequired,
    tabs: React.PropTypes.arrayOf(
        React.PropTypes.shape({
            label: React.PropTypes.string.isRequired,
            id: React.PropTypes.string.isRequired,
            key: React.PropTypes.number.isRequired
        })
    ).isRequired
}
```

And for the content

``` javascript
propTypes: {
    selectedTabKey: React.PropTypes.number.isRequired,
    tabs: React.PropTypes.arrayOf(
        React.PropTypes.shape({
            label: React.PropTypes.string.isRequired,
            content: React.PropTypes.element.isRequired,
            id: React.PropTypes.string.isRequired,
            key: React.PropTypes.number.isRequired
        })
    ).isRequired
}
```

robbie-c:
Then when a tab navigation is clicked, it calls `onTabSelected` which dispatches an action, which updates the store, which updates `selectedTabKey` that is passed to the content.

blocks/core/ff_module/ff_module-tabs-navigation-component/ff_module-tabs-navigation-component.js
```diff
@@ -0,0 +1,45 @@
...
+	onTabClick: function(event, tabKey) {
+		event.preventDefault();
+		this.props.onTabSelected(tabKey);
+	},
+	render: function() {
+		var {onTabSelected, tabs} = this.props;
+		return <ul className='ff_module-tabs-navigation'>
+			{tabs.map(function(tab) { 
+				return <li key={tab.key} className={generateClass('ff_module-tabs-navigation__tab', this.props) + isSelectedTab(tab.key, this.props.selectedTabKey)} data-ff_module-tabs-react-target={tab.id}>
+						<a href="#" onClick={(event) => this.onTabClick(event, tab.key) } className='ff_module-tabs-navigation__link' >{tab.label}</a>
```
dkhuntrods, line 41: Do you still need these attributes `data-ff_module-tabs-react-target`? If not, should probably remove for clarity.

blocks/core/ff_module/ff_module-tabs-navigation-component/_ff_module-tabs-navigation-component-renderer.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+var ReactDOM = require('react-dom');
+var createStore = require('redux').createStore;
+var ReactRedux = require('react-redux');
```
dkhuntrods, line 5: These (redux) are only being used in the renderer, right? Just checking in case if we need to update the externals in webpack. In general I think they should only be being used there...

sd-92, line 5: Yes they are only used in the renderers!

dkhuntrods:
Hi Simone, I think there are a few more unused data-attrs floating around - once you can get rid of those, LGTM.

PR #158, opened 2016-05-04
Add file list to formPreview

sd-92:


blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.js
```diff
@@ -14,6 +15,7 @@ module.exports = React.createClass({
 				value: React.PropTypes.string,
 				modifier: React.PropTypes.string,
 				previewFor: React.PropTypes.string,
+				fileList: React.PropTypes.object,
```
dkhuntrods: Might be a good idea to point to the propTypes definition in the FileList component itself here (with a `shape` wrapper) but not strictly necessary

dkhuntrods:
One small comment; otherwise lgtm

sd-92:
@dkhuntrods I've made the change :) if it looks good I'll go ahead and merge!

blocks/core/ff_module/ff_module-form-preview/ff_module-form-preview.js
```diff
@@ -14,6 +15,9 @@ module.exports = React.createClass({
 				value: React.PropTypes.string,
 				modifier: React.PropTypes.string,
 				previewFor: React.PropTypes.string,
+				fileList: React.PropTypes.shape({ 
+					files: React.PropTypes.array(FileList.propTypes.files)
```
dkhuntrods: Does `...shape(FileList.propTypes)` not work?

sd-92:
@dkhuntrods You're right sorry, that works and looks less messy!

dkhuntrods:
@Simonedeluca great! lgtm then :)

PR #159, opened 2016-05-04
Dkh task event states

dkhuntrods:
I want to begin merging this now; not all display text is 100% accurate, and there's probably some refactoring needed, but everything is functional and the rest is just minor tweaks.


blocks/core/ff_module/ff_module-task-event/_src/taskEventWithOptionalMessage.js
```diff
@@ -1,17 +1,67 @@
...
 
-        return  <TaskEventBase description={this.props.description}>
+        return  <TaskEventBase
+                    description={this.props.description}
+                    actions={this.props.actions}
+                    state={this.props.state}
+                    onRetryAfterStatusError={this.props.onRetryAfterStatusError}>
+                    <p className='ff_module-task-event__author-action'>{outputText}</p>
+                    {message}
+                </TaskEventBase>;
```
bwfox: Some duplication here between the rendering of the different states. Could extract this out into a function that takes `props`, `outputText` and `message` as parameters.

bwfox:
Left a comment about one minor thing. Other than that, LGTM.

blocks/core/ff_module/ff_module-task-event/_src/taskEventWithOptionalMessage.js
dkhuntrods: Yes, agreed, will tackle that. Thanks.

bwfox:
@dkhuntrods I am wondering if instead of having `Sent` and `Saved` states, we want just one of those, coupled with the `released` state. So `sent` would be `released`, `saved` would be `!released`. This would applied in a similar way for the `pending` and `error` states. E.g., `error` + `released` is `errorSending`, `error` + `!released` is `errorSaving`. Etc.

I think that would also make the front-end logic simpler as the pending, saved, error part of the state when saving is independent from the released/unreleased part.

dkhuntrods:
This is done; just waiting for webapp to merge the branch with supporting changes before merging into dev.

PR #167, opened 2016-05-10
Sd Responses screen - responsive patterns

sd-92:


dkhuntrods:
Hi Simone, I've had a look at this, and just want to check I'm seeing what you've intended.
Are you still working on this?

dkhuntrods:
Hiya Simone,
This is looking pretty good, however I think you'll need to modify the approach to have the components being passed in, rather than creating them from props. 
So, what I think I'd do is 
- Modify the existing task event list to have a 'currentTaskResponses' (or some similar name) prop; this would be a component and is what would be rendered below the currently selected item (doesn't need to know anything, just renders it)
- Then the task-response component would just accept 'list' (or whatever) and 'detail' props, which would be the updated task event list, and the current task responses component
- It needs to work with this file; as you see, lots of things are being created there that the app layer needs to know about, so this component can't create them from props https://github.com/fireflylearning/firefly-webapp/blob/5f71d76c046c3db8513c03fbffc48f9b618dbfc9/front-end/lib/task-responses/containers/Responses.js

blocks/core/ff_module/ff_module-response-recipient-list/ff_module-response-recipient-list.js
```diff
@@ -1,13 +1,34 @@
...
+				isRead ={response.isRead}
+				isSelected = {response.isSelected}
+				label = {response.label}
+				event = {response.latestEvent}
+				markAndGrade={response.markAndGrade}
+				lastEventWasAuthoredByCurrentUser={response.lastEventWasAuthoredByCurrentUser}
+				onSelect={()=>this.props.onSelect(response.recipient)}
+				pic_href={response.pic_href}
+				/>
+				{currentResponse}
```
dkhuntrods: I would say this should only render if the response is currently selected
So something like: response.isSelected? {currentResponse} : null

blocks/core/ff_module/ff_module-response-recipient-list/ff_module-response-recipient-list.js
```diff
@@ -1,13 +1,34 @@
...
     displayName: 'TaskResponseRecipientList',
     propTypes:{
         responses: React.PropTypes.array.isRequired,
         onSelect: React.PropTypes.func.isRequired
     },
-    render: template
+    render: function() {
+    	return <ItemRepeater modifier='separated'>
+    		{this.props.responses.map((response)=>{
+    			var currentResponse = response.currentTaskResponse ? <div className='ff_module-response-recipient-list__currentResponse'>{response.currentTaskResponse}</div> : '';
```
dkhuntrods: The currentResponse ought be a prop on the main component, not the responses - it's just a single item, like the 'details'.

blocks/core/ff_module/ff_module-profile-response-button/_ff_module-profile-response-button.rt.js
```diff
@@ -10,7 +10,7 @@ module.exports = function () {
         'className': 'ff_module-profile-response-button__image',
         'src': this.props.pic_href
     })), React.createElement('dl', { 'className': 'ff_module-profile-response-button__caption' }, React.createElement('dt', { 'className': 'ff_module-profile-response-button__heading' }, !this.props.lastEventWasAuthoredByCurrentUser ? React.createElement('span', { 'className': this.generateIconClass() }) : null, React.createElement('span', { 'className': 'ff_module-profile-response-button__label' }, this.props.label)), this.props.event ? React.createElement('dd', { 'className': 'ff_module-profile-response-button__status' }, React.createElement('span', { 'className': 'ff_module-profile-response-button__status-text' }, this.renderStatus()), this.props.event ? React.createElement('time', {
-        'dateTime': this.props.event.description.sent,
+        'dateTime': this.props.event.sent,
         'className': 'ff_module-profile-response-button__status-sent'
     }, ' ', this.renderTime()) : null) : null, this.props.markAndGrade ? React.createElement('dd', { 'className': 'ff_module-profile-response-button__mark-and-grade' }, this.renderMarkAndGrade()) : null));
 };
```
dkhuntrods: This file shouldn't be around anymore, maybe the merge will delete it

sd-92: Yeah that got deleted in the merge! 💃

blocks/core/ff_module/ff_module-response-recipient-list/ff_module-response-recipient-list.js
```diff
@@ -1,21 +1,21 @@
...
+var ProfileResponseButton = require('../ff_module-profile-response-button/ff_module-profile-response-button.js');
+var ItemRepeater = require('../../ff_container/ff_container-item-repeater/ff_container-item-repeater.js');
 
 module.exports = React.createClass({
     displayName: 'TaskResponseRecipientList',
     propTypes:{
         responses: React.PropTypes.array.isRequired,
-        onSelect: React.PropTypes.func.isRequired
+        onSelect: React.PropTypes.func.isRequired,
+        currentTaskResponse: React.PropTypes.element
```
dkhuntrods, line 12: Don't forget to update the `.md` too

blocks/core/ff_module/ff_module-response-recipient-list/_ff_module-response-recipient-list-renderer.js
```diff
@@ -10,7 +10,7 @@ var responses = [{
     },
     guid: "u47",
     label: "Sally Student",
-    latestEvent: {
```
dkhuntrods, line 13: actually, these should still be `latestEvent`, but the inner object needs to be wrapped in `description`;

```
latestEvent:{
   description: {
        type: eventTypes.markAndGrade,
        sent: new Date()
    }
}
```

blocks/core/ff_module/ff_module-responses/ff_module-responses.js
```diff
@@ -0,0 +1,20 @@
...
+		return <div className='ff_module-responses'>
+			<div className='ff_module-responses__recipients'>
+				{this.props.recipientList}
+			</div>
+			<div className='ff_module-responses__content'>
+				{this.props.responseDetails}
+			</div>
+		</div>
+	}
+});
```
dkhuntrods: How is this file's functionality distinct from `ff_container-scrollable-list`? Is this file necessary?

blocks/melody/ff_container/ff_container-scrollable-list/ff_container-scrollable-list.less
```diff
@@ -1,12 +1,12 @@
...
+	width: 100%;
+	border-collapse: collapse;
+	border:2px solid @ff_colour_control_border;
+	height: 585px;
+	flex-direction: column;
     align-items: stretch;
-    box-sizing: border-box;
     &__item {
 
     }
```
dkhuntrods: Something in the changes here (either less or config, not sure) seems to have broken things, look at:
http://localhost:3000/blocks/core/lib_test/lib_test-task-responses-screen/lib_test-task-responses-screen.html

dkhuntrods:
@simonedeluca-firefly  A few comments for you; as a general rule, do a quick search to make sure that you're updating components whereever they might be used (ie, within other modules, in lib_test, etc)

blocks/core/ff_module/ff_module-response-recipient-list/ff_module-response-recipient-list.js
```diff
@@ -1,21 +1,21 @@
...
     },
     render: function(){
-        return (
-            <ItemRepeater modifier="separated">
-                {this.props.responses && this.props.responses.map(response =>
-                    <ProfileResponseButton
+        return <ItemRepeater modifier="separated">
+                {this.props.responses && this.props.responses.map(response =>{
+                	var currentResponse = response.isSelected ? <div className='ff_module-response-recipient-list__currentResponse'>{this.props.currentTaskResponse}</div> : null;
+                    return <div><ProfileResponseButton
```
dkhuntrods: On line 18, the `div` will need a key as it's inside a map...you could take the key from line 19 and use it here instead, as it's the outer item now.

blocks/core/ff_module/ff_module-responses/ff_module-responses.js
sd-92: They are basically the same! So I think what I should just is just move the .less styles(of the responses module) in the scrollable list to make it responsive. I don't think this ff_module-responses is needed anymore. I created it thinking it would do something different but it just added more complexity.

dkhuntrods:
Hey Simone! Looking really good, I think we're nearly there. Just css tweaks now mainly:
- I think the breakpoint for expanding the scrollable list needs to be a bit wider; you can see there are some strange effects at the narrower size with the two columns. In the designs, it doesn't seem to get wide until around 960, so could look at that.
- The TaskMetaActions might need some css tweaks at small sizes to align with the designs more closely; the TaskResponseActions might need it too, if changing the breakpoint doesn't help.
- Think the `ff_module-response-recipient-list.md` file could use updating, as well as the renderer; the currentTaskResponse doesn't seem to be rendering now.
- Don't forget to delete `ff_module-responses` if you're not using it anywhere now.

blocks/core/ff_module/ff_module-task-meta-actions/ff_module-task-meta-action.less
```diff
@@ -0,0 +1,21 @@
...
+				order: 1;
+			}
+			order: 2;
+		}
+		.ff_container-control-bar__group:last-of-type {
+			@media (min-width: @ff_bp_wide) {
+				justify-content: flex-end;
+				order: 2;
+			}
+			order: 1
```
dkhuntrods: The missing semi-colon is throwing an error on my machine

dkhuntrods:
Ok! So close, but there's a final couple tweaks:
- Instead of nesting the css rules for `container-control-bar-split` inside `task-meta-actions`, pass in a new `classes` prop that indicates that task-meta-actions is the owner (this situation is what the 'classes' prop is for), so in the js:
  `<ContainerControlBar modifier="split" classes="ff_module-task-meta-actions__controlbar">`
  and then refer to that in the css.
- The upper margin on the task-responses component (http://localhost:3000/blocks/core/ff_module/ff_module-task-responses/ff_module-task-responses.html) is now a bit too big; it's a bit too big when it's in the container as well: http://localhost:3000/blocks/core/lib_test/lib_test-task-responses-screen/lib_test-task-responses-screen.html
- The buttons on the task-responses component should only disappear when they're inside the container, currently they disappear in the standalone module too: http://localhost:3000/blocks/core/ff_module/ff_module-task-responses/ff_module-task-responses.html. If necessary, add a 'classes' prop here too.

dkhuntrods:
Good to go!

PR #168, opened 2016-05-10
Prevent NaN errors in date picker related input fields

boonier:


boonier:
@dkhuntrods all good?

PR #170, opened 2016-05-10
Dkh task global states

dkhuntrods:


blocks/core/ff_module/ff_module-online-status/ff_module-online-status.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var React = require('react');
+
+module.exports = React.createClass({
+    displayName: 'OnlineStatus',
+    propTypes: {
+        state: React.PropTypes.shape({
+            offline: React.PropTypes.bool,
+            online: React.PropTypes.bool
```
bwfox: I'd just have one of these, otherwise you leave the door open for weird states where the component is both online and offline.

dkhuntrods: Right! ok, will remove one.

tests/ff_module-task-meta-actions.test.js
```diff
@@ -0,0 +1,86 @@
...
+            id: 'filter-3'
+        }]
+    },
+    singleActions: [{ text: '[Send All Now]', onClick: function() {} }],
+    groupedActions: {
+        text: '[More Actions]',
+        list: [{ text: '[Edit]', onClick: function(event) { console.log(event); } }]
+    }
+}, {
+    state: {
```
bwfox, line 44: There's another state object on line 15 in this file. Are both of these intended to exist and if so what's the difference between them?

bwfox:
Couple of comments, other than that LGTM.

tests/ff_module-task-meta-actions.test.js
dkhuntrods, line 44: Yeah, it's an array of props, the first tests the full range, the second tests only that the archive state works, so doesn't need all the other values

dkhuntrods:
Great, thanks! - replied inline

PR #175, opened 2016-05-17
ff_module-formsteps responds to url hashes

boonier:


boonier:
@dkhuntrods could you check pleeeease

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
```diff
@@ -91,12 +90,38 @@ function getTabHandler($root, options) {
...
+
+    function getHash() {
+        var hash = window.location.hash.substring(1);
+        if (hash) return hash;
+        return null;
+    }
+
+    function getIndexOfTrigger(hash) {
+        var $triggers = $root.find(linkSel);
+        var index = $($triggers).filter(function(i) {
```
dkhuntrods: $triggers should be a jQuery object already, so you shouldn't have to wrap it again to filter? I could be wrong, but worth checking to avoid an additional instantiation.

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
```diff
@@ -122,12 +147,12 @@ function getTabHandler($root, options) {
...
             $lastContent = removeActiveClasses($activeContent);
             addVisitedClasses($lastLinks, $lastContent);
-            addActiveClasses($selectedLinks, $selectedContent);
+            addActiveClasses($selectedLinks, $selectedContent, index);
             visitedCallback($lastLinks, $lastContent, $selectedLinks, $selectedContent);
-
             if (isComplete) {
                 addCompleteClasses($lastLinks, $lastContent);
                 completeCallback($lastLinks, $lastContent, $selectedLinks, $selectedContent);
+                setHash(target);
```
dkhuntrods: The hash should be set on `canAdvance` rather than `isComplete`

blocks/core/_lib/ff-tabs/_ff-tabs-core.js
```diff
@@ -91,12 +90,41 @@ function getTabHandler($root, options) {
 
     function init() {
         $root.on('click', linkSel, handleClick);
-        setInitialTab();
+        var index = getHashIndex() || 0;
+        setActiveTab(index);
```
dkhuntrods, line 94: should update selectedIndex here I'd say

PR #178, opened 2016-05-20
First few bits for Nautilus pattern library

aegirfirefly:
Form steps
Header
Tabs
Buttons


blocks/nautilus/ff_container/ff_container-dashboard-item-repeater/ff_container-dashboard-item-repeater.less
```diff
@@ -0,0 +1,11 @@
+.ff_container-dashboard-item-repeater__items {
```
roobottom, line 1: You could use the style 

```
.ff_container-dashboard-item-repeater {
   &__items {}
   &__item {}
}
```

blocks/nautilus/ff_container/ff_container-dialog/ff_container-dialog.less
```diff
@@ -0,0 +1,31 @@
...
+        color: @ff_colour_control_text;
+        margin: 0;
+        padding: @ff_size_control_vertical @ff_size_control_horizontal;
+        position: relative;
+    }
+    &__title {
+        color: @ff_colour_control_text;
+    }
+    &__close-top {
+        #ff_util > .ir;
```
roobottom, line 14: Doesn't this need to be `#ff_util > .ir()` ?

blocks/nautilus/ff_container/ff_container-filter/ff_container-filter.less
```diff
@@ -0,0 +1,13 @@
...
+	border-bottom: 1px solid @ff_colour_control_border; // outer borders to be added by repeater/container module
+}
+.ff_container-filter__label {
+	display: inline-block;
+	cursor: inherit;
+}
+.ff_container-filter__items {
+	background: lighten(@ff_colour_control_bg_selected,15%);
+	padding: @ff_size_control_vertical;
+}
```
roobottom, line 13: This could be in the style

```
.ff_container-filter {
    &__heading {}
}
```

roobottom:
I've added a few inline comments.

My main comment is: Are these all just copied from Melody's styles? I'm not convinced by this approach, as it feel like you could miss things. Isn't it better to start with a blank file for each pattern, rather than copying all the styles in at the start? Sure, you could copy bits of them as you go along, but that feels like you're more in control that way.

PR #181, opened 2016-05-20
Pagination!

aegirfirefly:


blocks/nautilus/ff_module/ff_module-pagination/ff_module-pagination.less
```diff
@@ -0,0 +1,16 @@
+.ff_module-pagination__items {
```
roobottom: I think we prefer the style `&__items`

roobottom:
LGTM

PR #182, opened 2016-05-23
Two modules, but while doing those noticed a couple of bugs

aegirfirefly:
Two PRs got combined:
Progress bars
Segmented control
Class views, and planner grid (showing the classes)

Fixed while doing the above:
Missing bits in buttons
Removed (bug) unnecessary style in file list


roobottom:
LGTM

PR #184, opened 2016-05-24
Dkh react redux forms

dkhuntrods:


blocks/core/ff_container/ff_container-form-line/ff_container-form-line.js
```diff
@@ -1,36 +1,72 @@
...
+
+    if (React.isValidElement(node)) {
+        return React.cloneElement(node, { className: nodeClass });
+    } else {
+        return <span className={nodeClass}>{node}</span>;
+    }
+}
+
+function getNodeName(node){
+    return node.displayName || node.type.displayName || node.type.name || node.type;
```
bwfox, line 24: Just wondering why there are four possible places the node name might come from?

dkhuntrods, line 24: That's just React! The constructor can be a simple function, or a class, and the name can live in different places on either of those.

bwfox:
LGTM

PR #194, opened 2016-06-02
Dkh refactor task editor

dkhuntrods:
Refactoring TaskEventEditor to allow for varying text & buttons


tests/ff_module-task-event-editor.test.js
```diff
@@ -234,5 +247,71 @@ describe('TaskEventEditor', function() {
...
+                if (spyName === 'onClose') {
+                    expect(spies[spyName].calledTwice).to.equal(true);
+                } else  {
+                    expect(spies[spyName].calledOnce).to.equal(true);
+                }
+
+            })
+
+
+        });
```
bwfox, line 323: It might be worth extracting some parts of this test out into functions in order to keep the length down. I think that'll make it easier to read.

bwfox:
One comment, other than that LGTM.

PR #196, opened 2016-06-08
Update Task Event Status styles

dkhuntrods:
Updates to make the task event status appear less 'buttony'

### DIFF
--- a/blocks/core/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
+++ b/blocks/core/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
@@ -2,20 +2,17 @@
 
     display: inline-block;
     font-size: .875rem;
-    padding: @ff_size_spacing_min*0.25 @ff_size_spacing_min;
+
     // conceptually if not actually a child
     &__transition-wrapper {
         position: relative;
     }
 
     &__icon {
-        margin: -.4em .4em -.4em -.25em;
+        margin-right: 0.25rem;
     }
 
     button&__error-button {
         font-size: .875rem;
     }
-
-
-
 }
--- a/blocks/melody/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
+++ b/blocks/melody/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
@@ -1,19 +1,20 @@
+@base-height: 1.25rem;
+@visible: 1;
+@invisible: 0.01;
+
 .ff_module-task-event-status {
 
     color: @ff_colour_text_quiet;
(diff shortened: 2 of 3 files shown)


blocks/core/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
```diff
@@ -2,20 +2,17 @@
...
-    padding: @ff_size_spacing_min*0.25 @ff_size_spacing_min;
+
     // conceptually if not actually a child
     &__transition-wrapper {
         position: relative;
     }
 
     &__icon {
-        margin: -.4em .4em -.4em -.25em;
+        margin-right: 0.25rem;
```
roobottom, line 12: Is this layout? and if so, should it be in the core at all?

blocks/melody/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
```diff
@@ -1,19 +1,26 @@
+@base-height: 1.25rem;
+@visible: 1;
+@invisible: 0.01;
+
 .ff_module-task-event-status {
 
+    font-size: .8rem;
```
roobottom: Isn't it better to leverage less's power to calculate sizes like this? E.g., `font-size: unit(20/@base,rem);` ?

blocks/core/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
dkhuntrods, line 12: It's layout but it's pretty fundamental and most themes won't have to override this, in which case I don't think it's necessary to add it to every theme when it can appear here.

blocks/melody/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
dkhuntrods: If we were using the vars in a really consistent way across the designs such that changing one at the base of the theme changed them consistently everywhere, then yes, but we're not really doing that so I just went for simplicity. Nonetheless, if you'd prefer it, no prob.

roobottom: OK, I was just thinking for others coming in and reading this code, it makes it easier to see the original values you intended if they don't have to do the mental `times @base` calculations.

blocks/core/ff_module/ff_module-task-event-status/ff_module-task-event-status.less
roobottom, line 12: OK, fair enough. 👍

PR #198, opened 2016-06-10
Build of Folio for Epic 1 components only

aegirfirefly:
Dashboard section, landing section and headings/links - visible in FF5 with FF6 nav.

Also has settings and colours from Folio - the rest of the patterns currently apply these wrong, so please ignore.


dkhuntrods:
All I can see are icon file changes, is that right?

aegirfirefly:
No, there are changes to the containers listed in the PR message. The icons are just from copying them over.

dkhuntrods:
Would those be .xsl files? I can't see them in the diff :<

aegirfirefly:
They are there, right after all the icons. Github complains it can't show the whole diff because too many files have changed.

dkhuntrods:
Ok, even after checking out I'm still not seeing any .xsl files!

dkhuntrods:
Right, can't leave comments inline - 

```
    &__main {
        #ff_mix-sections > .main();
        padding: 0 @ff_size_control_horizontal;
    }
```

Since that padding is added in both sections, would it make sense to add it into `#ff_mix-sections > .main();`?

Otherwise, LGTM

aegirfirefly:
@dkhuntrods I can't be sure until later - if I find myself adding it to other places where it's used I'll add it to the mixin.

PR #200, opened 2016-06-13
Add invalid event type component in case something goes wrong.

bwfox:
Unlikely to happen in production, but can be fairly common under
development; previously the page would break.


dkhuntrods:
LGTM

PR #205, opened 2016-06-20
Increase hit area of columnar list item

boonier:
Adds an extra span wrapper around content to control the spacing consistently for both Links and Non-links.

Also corrected the md file - 'two-column' modifier was spelt incorrectly.


aegirfirefly:
The Melody fix looks fine, can you do the same for Folio and Nautilus - or in core?

boonier:
Hmm yes it does seem like a similar scenario that might be best place in the core style. @dkhuntrods any objections?

dkhuntrods:
@aegirfirefly @boonier I can't help but feel that in this case it would be preferable not to have to have the span - if the link is inline-block, does the padding apply to both that and the text content?
It just seems like there should be a simpler way to do that...

boonier:
The reason for that span is if there is no link `<a>` on the list item, then there is no hook to style it with, and the padding needs to reside within the `<a>` not the parent `<li>` to increase the hit area. If only the links have the vertical padding then spacing of the items goes out of wack.

If it is unlikely that the contents of a columnar-list is going to be anything but a link, then we could dispense of the span, sure.

dkhuntrods:
What about __item > \* {} for a hook?

boonier:
Have updated the PR that removes the dependancy on span markup. The above __item > \* {} would only  affect a child afaik (unless I'm missing something), and the list item without url just contains the raw text content; no child node. So I've gone down the modifier route and it's working well.

aegirfirefly:
Cool. Could we have it so that a span only gets added if there's no link? I assume we'll have columnar lists in variable non-linky places.

boonier:
What I've done doesn't need the span tag and is only dependant on the class modifier. `ff_module-columnar-list--no-url` is only added if there is no presence of a url in the xml.

dkhuntrods:
Ok...think that's a bit cleaner; I'd probably invert the check myself as you're already checking for a url and could just hook into that (so the class would only be added if url present, and class would be `--with-url` or something). To my mind, the li is the standard case, and having a url is an addition or modifier.

boonier:
@dkhuntrods cool good point

Re the original question, should this be part of the core styles?

boonier:
Updated. Should I remove the empty theme specific files (melody, nautilus, folio)?

dkhuntrods:
LGTM, if @aegirfirefly's happy too :)

PR #206, opened 2016-06-23
Task Details Individual (Student and Parent)

dkhuntrods:
**Please don't merge**
Create Task Details Screen (Student/Individual) test page, and update components as required.
There is no component for this screen, as (following Task Details > Responses) it makes more sense as a webapp container.
Also adds a generic Modal with Dialog component, to simplify implementation where both are used as standard.
See http://localhost:3000/blocks/core/lib_test/lib_test-task-responses-screen-individual/lib_test-task-responses-screen-individual.html for an implementation of all the components required.


blocks/core/ff_module/ff_module-task-event/_src/MarkAndGradeTaskEvent.js
```diff
@@ -41,19 +40,18 @@ function getEditedFlag(props) {
...
-function getMessage(props) {
-    var description = props.description,
-        messageText = description.message,
-        editedFlag = getEditedFlag(props, 'messageEdited');
+function TaskEventMessage(props) {
+    var description = props.description || {},
+        messageText = description.message;
 
-    return messageText ? <p className="ff_module-task-event__message">{messageText}</p> : null;
+    return messageText ? <p className="ff_module-task-event__message">{messageText}</p> : <span/>;
```
robbie-c, line 47: Does null not work here? I kind of assumed that react could deal with this.

robbie-c:
I don't think I can review the CSS, but the JS LGTM. This was a pretty big PR.

dkhuntrods:
@robbie-c thanks, and apologies, it did end up that way.

sd-92:
CSS LGTM!

PR #213, opened 2016-06-30
Add fix for IE10 failure to deal with css columns and inline-block

boonier:


blocks/core/ff_module/ff_module-columnar-list/ff_module-columnar-list.less
```diff
@@ -24,6 +24,9 @@
 		width: 100%;
 		padding: @ff_size_control_vertical/2 0;
 	}
+	.ie10 &__item {
```
dkhuntrods, line 32: We don't have browser targeting prefixes in the patternlib yet, and have been trying to avoid having to add them - is there any other way this could be solved without targeting?

dkhuntrods:
Can you leave a NOTE for the time being, that there are no browser targeting prefixes as yet in patternlib, so this will target code on webapp? And a TODO to add ie10 target.

PR #214, opened 2016-07-05
ff_container-dialog: Add arrow modifier

boonier:


dkhuntrods:
LGTM! Don't forget to update webapp after merging :)

